PR #20, opened 2022-04-20
implement GHASH 2PC

themighty1:


tls-core/src/ghash/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+//! computed by R locally. Only terms 2 and 3 will be computed using
+//! GHASH 2PC. R will obliviously request values for bits of H1_r and H2_r.
+//! The XOR sum of all values which S will send back plus H1_r*H2_r will
+//! become R's share of H3.
+//!
+//! When performing block multiplication in 2PC, Master holds the Y value and
+//! Slave holds the X value. The Slave then computes the X table and masks it.
+
+mod common;
+mod errors;
```
sinui0, line 21: The errors module should be public so users can import and design their control flow around them

tls-core/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 #[cfg(feature = "prf")]
 pub mod prf;
+pub mod ghash;
```
sinui0, line 3: Let's add a feature gate for the ghash module, same as how I had the prf module configured.

Also we should add a newline at the end of this file

sinui0 (changes requested):
2 comments, otherwise lgtm

PR #22, opened 2022-04-24
beaver derandomization for OT extension

sinui0:
I've implemented a supertrait over top of ExtSendCore and ExtReceiveCore to allow for Beaver Derandomization


themighty1:
lgtm

PR #24, opened 2022-04-30
use enum for prf

themighty1:


tls-core/src/prf/master.rs
```diff
@@ -1,225 +1,179 @@
...
 pub struct MasterKe3 {
     /// H((ms xor ipad) || a1 || seed)
     pub inner_hash_p1: [u8; 32],
     /// H((ms xor ipad) || a2 || seed)
     pub inner_hash_p2: [u8; 32],
 }
 
-impl PrfMaster<Initialized> {
+#[derive(std::cmp::PartialEq, Copy, Clone)]
+pub enum PRFMasterState {
```
sinui0, line 55: This state will be mostly private (though we may want to expose it via a method), it's safe to be more succint. Eg. the enum can be named `State` and you don't need to prefix the variants. e.g
```
pub enum State {
  Initialized,
  Ms1,
  Ms2,
  Ms3
  ...
}
```

tls-core/src/prf/master.rs
```diff
@@ -1,225 +1,179 @@
...
+    MasterMs1Sent,
+    MasterMs2Sent,
+    MasterMs3Sent,
+    MasterKe1Sent,
+    MasterKe2Sent,
+    MasterKe3Sent,
+}
+
+#[derive(Copy, Clone)]
+pub struct PrfMaster {
```
sinui0, line 66: I'm not sure why I originally named this PrfMaster, PRFMaster is preferable

tls-core/src/prf/mod.rs
```diff
@@ -6,9 +7,32 @@ mod utils;
...
 pub use slave::PrfSlave;
 
+use master::{
+    KeInit, MasterKe1, MasterKe2, MasterKe3, MasterMs1, MasterMs2, MasterMs3, MasterSetup,
+};
+use slave::{SlaveKe1, SlaveKe2, SlaveMs1, SlaveMs2, SlaveMs3};
+
+#[derive(Copy, Clone)]
+pub enum PRFMessage {
+    MasterSetup(MasterSetup),
```
sinui0, line 17: This isn't a message, we should pull it out and pass it in either the constructor or a standalone `fn setup(...)` method

tls-core/src/prf/slave.rs
```diff
@@ -1,131 +1,117 @@
...
-
-        (
-            SlaveMs1 { a1 },
-            PrfSlave {
-                state: Ms1 { outer_hash_state },
-            },
-        )
-    }
+#[derive(std::cmp::PartialEq, Copy, Clone)]
+pub enum PRFSlaveState {
```
sinui0, line 36: Same comments as for Master's state

tls-core/src/prf/slave.rs
```diff
@@ -1,131 +1,117 @@
...
-impl PrfSlave<Ke2> {
-    pub fn next(self, m: MasterKe2) -> (SlaveKe2, PrfSlave<Cf>) {
-        // H((pms xor opad) || H((pms xor ipad) || a1))
-        let a2 = finalize_sha256_digest(self.state.outer_hash_state.clone(), 64, &m.inner_hash);
 
-        (SlaveKe2 { a2 }, PrfSlave { state: Cf })
+    pub fn next(
+        &mut self,
+        message: PRFMessage,
+        outer_hash_state: Option<[u32; 8]>,
```
sinui0, line 64: Having this option in the function signature is clunky. Can we pass it in another way?

sinui0 (changes requested):
Structurally, it looks good and is what I had in mind. I'm undecided, but perhaps we can improve readability by implementing the various steps in private methods called within the next() match block. Easier to write unit tests as well.

tls-core/src/prf/slave.rs
```diff
@@ -1,131 +1,159 @@
...
 }
 
+#[derive(Copy, Clone)]
 pub struct SlaveKe2 {
     /// H((ms xor opad) || H((ms xor ipad) || a1))
     pub a2: [u8; 32],
 }
 
-pub struct SlaveCf {}
+#[derive(std::cmp::PartialEq, Copy, Clone)]
```
sinui0, line 35: PartialEq should be in scope and doesn't require std::cmp::

tls-core/src/prf/mod.rs
```diff
@@ -70,34 +103,64 @@ mod tests {
...
 
-        let p1 = finalize_sha256_digest(outer_hash_state.clone(), 64, &inner_hashes.inner_hash_p1);
-        let p2 = finalize_sha256_digest(outer_hash_state, 64, &inner_hashes.inner_hash_p2);
+        let p1 = finalize_sha256_digest(outer_hash_state, 64, &inner_hash_p1);
+        let p2 = finalize_sha256_digest(outer_hash_state, 64, &inner_hash_p2);
 
         let mut ek = [0u8; 40];
         ek[..32].copy_from_slice(&p1);
         ek[32..].copy_from_slice(&p2[..8]);
+        let mut cwk = [0u8; 16];
```
sinui0, line 145: what is this used for?

tls-core/src/prf/mod.rs
```diff
@@ -70,34 +103,64 @@ mod tests {
...
         let client_write_key = &ek[..16];
         let server_write_key = &ek[16..32];
         let client_write_iv = &ek[32..36];
         let server_write_iv = &ek[36..];
+
+        // Seems like the simplest way to compare byte slices is to hex encode
+        // them first.
+        assert_eq!(
+            hex::encode(&client_write_key),
+            "24542375e0909cabc9976eaa009d283a"
```
sinui0, line 157: Where did these static values come from?

sinui0 (commented):
Couple comments, otherwise lgtm

tls-core/src/prf/mod.rs
themighty1: Those are the result on running this very test_prf() on the inputs. Should we pull in rustls as a dev dep and run the PRF with it to compare the output?

tls-core/src/prf/mod.rs
themighty1: Not in use, removed.

tls-core/src/prf/mod.rs
themighty1: I think instead of rustls Im gonna compare our output to this https://scapy.readthedocs.io/en/latest/api/scapy.layers.tls.crypto.prf.html

sinui0: As long as we check them against some other known working implementation we should be good. Either pulling in a dependency or pasting in is fine with me.

PR #29, opened 2022-05-20
KOS15 check

themighty1:


themighty1:
I didn't know what names to give in tests in /src/proto/core/ot.rs , so I just assigned tmp names .

mpc-core/src/utils.rs
```diff
@@ -65,6 +66,18 @@ pub fn parse_ristretto_key(b: Vec<u8>) -> Result<RistrettoPoint, std::io::Error>
...
 }
 
+pub fn sha256(data: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    hasher.update(data);
+    hasher.finalize().into()
+}
+
+/// XORs 2 slices of bytes of equal length
+pub fn xor(a: &[u8], b: &[u8]) -> Vec<u8> {
```
sinui0, line 76: This always results in a heap allocation (Vec) even if the iterables are on the stack. We could instead pass in a mutable reference:

```rust
fn xor(a: &[u8], b: &[u8], out: &mut [u8]) {
    for ((a, b), out) in a.iter().zip(b.iter()).zip(out.iter_mut()) {
        *out = a ^ b;
    }
}
```

Or we could alternatively avoid any copies/allocation

```rust
fn xor_replace(a: &mut [u8], b: &[u8]) {
    for (a, b) in a.iter_mut().zip(b.iter()) {
        *a ^= b;
    }
}
```

mpc-core/src/ot/extension/sender.rs
```diff
@@ -186,7 +222,45 @@ where
...
+        let mut qs = utils::transpose(&qs);
+
+        // Check correlation
+        // The check is explaned in the KOS15 paper in a paragraph on page 8
+        // starting with "To carry out the check..."
+        // We use the exact same notation as the paper.
+
+        // Seeding with a value from cointoss so that neither party could influence
+        // the randomness
+        let mut rng = ChaCha12Rng::from_seed(self.cointoss_random.unwrap());
```
sinui0, line 234: Avoid unwrap, will cause a panic

mpc-core/src/ot/extension/sender.rs
```diff
@@ -186,7 +222,45 @@ where
...
+        let mut check0 = U64x2::from([0u8; 16]);
+        let mut check1 = U64x2::from([0u8; 16]);
+        for j in 0..ncols {
+            let q = U64x2::from(&qs[j]);
+            // chi is the random weight
+            let chi = U64x2::random(&mut rng);
+            // multiplication in the finite field (p.14 Implementation Optimizations.
+            // suggests that it can be done without reduction).
+            let (tmp0, tmp1) = q * chi;
+            check0 = check0 ^ tmp0;
```
sinui0, line 245: check0 ^= tmp0

sinui0 (changes requested):
Just a few odds and ends, but lgtm

mpc-core/src/ot/extension/u64x2.rs
```diff
@@ -0,0 +1,199 @@
+use core::num::Wrapping;
```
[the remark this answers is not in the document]
sinui0: Here I'm referring to the u64x2 module.

mpc-core/src/proto/core/ot.rs
```diff
@@ -149,16 +167,28 @@ impl From<ot::ExtReceiverSetup> for ExtReceiverSetup {
...
             ncols: s.ncols as u32,
             table: s.table,
+            x: s.x.to_vec(),
+            t0: s.t0.to_vec(),
+            t1: s.t1.to_vec(),
         }
     }
 }
 
 impl From<ExtReceiverSetup> for ot::ExtReceiverSetup {
```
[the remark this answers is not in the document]
themighty1: I agree. Just to make sure, @sinui0 , the section of the code you highlighted is not what needs to be changed, because the direction there is core model -> prost model, right?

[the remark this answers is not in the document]
sinui0: `impl From<ExtReceiverSetup> for ot::ExtReceiverSetup` I believe this is prost -> core.

I'm _definitely_ going to refactor this after this PR to make it less confusing.

themighty1:
The clmul crate added. I tested it for both 32/64-bit soft backend as well as on an arm machine with features = "armv8"

clmul/src/backend/autodetect.rs
```diff
@@ -0,0 +1,130 @@
...
+cpufeatures::new!(mul_intrinsics, "aes"); // `aes` implies PMULL
+
+#[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
+cpufeatures::new!(mul_intrinsics, "pclmulqdq");
+
+/// Carryless multiplication
+#[derive(Clone, Copy)]
+pub struct Clmul {
+    // has_intrinsics is a flag that intrinsics are available
+    has_intrinsics: bool,
```
rozbb, line 23: No need for this flag. This is true iff `intrinsics` is `Some`

clmul/src/backend/autodetect.rs
```diff
@@ -0,0 +1,130 @@
...
+                has_intrinsics,
+                intrinsics: None,
+                soft: Some(soft::Clmul::new(h)),
+            }
+        }
+    }
+
+    pub fn clmul(self, x: Self) -> (Self, Self) {
+        if self.has_intrinsics {
+            let (r0, r1) = self.intrinsics.unwrap().clmul(x.intrinsics.unwrap());
```
rozbb, line 48: Maybe you can ignore this case, but what happens when you try to `clmul` a `soft` value with a `intrinsic` value? I believe it panics

clmul/src/backend/autodetect.rs
```diff
@@ -0,0 +1,130 @@
...
+                },
+            )
+        }
+    }
+}
+
+impl From<Clmul> for [u8; 16] {
+    #[inline]
+    fn from(m: Clmul) -> [u8; 16] {
+        if m.has_intrinsics {
```
rozbb, line 82: Repeating myself, but it's considered cleaner to use pattern matching. Like `match self.intrinsics` or `if let Some(in) = self.intrinsics`. The general heuristic is that you should rarely have to use `unwrap()` in real code

clmul/src/backend/pmull.rs
```diff
@@ -0,0 +1,124 @@
...
+#[inline(always)]
+unsafe fn pmull<const A_LANE: i32, const B_LANE: i32>(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t {
+    mem::transmute(vmull_p64(
+        vgetq_lane_u64(vreinterpretq_u64_u8(a), A_LANE),
+        vgetq_lane_u64(vreinterpretq_u64_u8(b), B_LANE),
+    ))
+}
+
+#[test]
+fn clmul_xor_eq() {
```
rozbb, line 99: You should move this test to `lib.rs` and delete the other identical test. No point in repeating these if only 1 of these will run anyway

clmul/src/backend/soft32.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/// Bit-reverse a 32-bit word in constant time.
+fn rev32(mut x: u32) -> u32 {
+    x = ((x & 0x5555_5555) << 1) | (x >> 1 & 0x5555_5555);
+    x = ((x & 0x3333_3333) << 2) | (x >> 2 & 0x3333_3333);
+    x = ((x & 0x0f0f_0f0f) << 4) | (x >> 4 & 0x0f0f_0f0f);
+    x = ((x & 0x00ff_00ff) << 8) | (x >> 8 & 0x00ff_00ff);
+    (x << 16) | (x >> 16)
+}
```
rozbb, line 213: Another reason to move the above tests to `lib.rs`, this file has no unit tests.

clmul/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+        let b: [u8; 16] = rng.gen();
+
+        let (r64_0, r64_1) = s64::new(&a).clmul(s64::new(&b));
+        let (r32_0, r32_1) = s32::new(&a).clmul(s32::new(&b));
+        assert_eq!(u128::from(r64_0), u128::from(r32_0));
+        assert_eq!(u128::from(r64_1), u128::from(r32_1));
+
+        cfg_if! {
+            if #[cfg(all(any(target_arch = "x86_64", target_arch = "x86")))] {
+                use clmul::Clmul as clm;
```
rozbb, line 115: Why are you branching on the backend in `lib.rs`? If you import `backend::Clmul` it should just do the right thing

mpc-core/src/ot/extension/mod.rs
```diff
@@ -202,6 +239,37 @@ mod tests {
...
+    // Test that the cointoss check fails on wrong data
+    fn test_ext_ot_cointoss_failure(mut sender: ExtSenderCore, mut receiver: ExtReceiverCore) {
+        use super::{ExtReceiveCore, ExtSendCore};
+
+        let mut base_sender_setup = receiver.base_setup().unwrap();
+        base_sender_setup.cointoss_commit = [77u8; 32];
+        let base_receiver_setup = sender.base_setup(base_sender_setup).unwrap();
+        let send_seeds = receiver.base_send(base_receiver_setup).unwrap();
+        let err = sender.base_receive(send_seeds).err().unwrap();
+        assert_eq!(err, ExtSenderCoreError::CommitmentCheckFailed);
```
rozbb, line 252: More idiomatically, here and in the rest of the tests:
```rust
let res = sender.base_receive(send_seeds);
assert_eq!(res, Err(ExtSenderCoreError::CommitmentCheckFailed)));
```
This will also produce a nicer failure message when `res` is `Ok`

mpc-core/src/ot/extension/mod.rs
```diff
@@ -229,6 +297,15 @@ mod tests {
         assert!(sender.is_complete());
         assert!(receiver.is_complete());
 
+        // Trying to send/receive more OTs should return an error
+        let err = sender.send(&[[Block::random(&mut rng); 2]]).err().unwrap();
```
rozbb, line 301: The array expression here probably doesn't do what you think it does. It creates a random `Block` X and makes an array `[X, X]`. You probably want `[X, Y]` for a uniform Y.

mpc-core/src/ot/extension/sender.rs
```diff
@@ -154,21 +175,47 @@ where
...
+        if self.state != State::BaseSetup {
+            return Err(ExtSenderCoreError::WrongState);
+        }
+        let receive = self.base.receive(payload.payload)?;
         self.set_seeds(receive);
-        self.state = State::BaseSetup;
+
+        // check the decommitment for the other party's share
+        if sha256(&payload.cointoss_share)
+            != self
```
rozbb, line 200: This should probably be a constant time equality check, no? Use `subtle::ConstantTimeEq` for this. The expression should be something like
```rust
if !bool::from(sha256(&payload.cointoss_share).ct_eq(other_value)) {
   // ...
}
```

mpc-core/src/ot/extension/sender.rs
```diff
@@ -186,7 +233,64 @@ where
...
+        let t0 = Clmul::new(&t0);
+
+        let mut t1: [u8; 16] = [0u8; 16];
+        t1.copy_from_slice(&receiver_setup.t1);
+        let t1 = Clmul::new(&t1);
+
+        let (tmp0, tmp1) = x.clmul(delta);
+        check0 ^= tmp0;
+        check1 ^= tmp1;
+        if !(check0 == t0 && check1 == t1) {
```
rozbb, line 286: Probably also `ct_eq`

mpc-core/src/ot/extension/sender.rs
rozbb: You should check the rest of the code for non-constant-time equalities. I'll probably miss some

rozbb (commented):
Nice work! Left some comments. Most are stylistic. Three things

1. You need to add attributes to the CLMUL files
2. There's not a lot of comments at all. I might be extreme, but I firmly believe that every single function, public or not, needs to have a comment describing what it does (boilerplate trait defs can be excluded tho). Comments on expressions are also great.
3. I'm not at all familiar with this protocol, so you have to use your judgement on whether or not the unit tests here cover enough. If you think so, then great.

sinui0:
@themighty1 I think we can simplify clmul considerably. We do not need the Clmul wrapper that you wrote which does runtime checking on whether intrinsics are available. That is something that is determined at _build_ time. All backend implementations provide the same interface and a type called Clmul.

I made an example branch of how this can be done instead: https://github.com/tlsnotary/tlsn/compare/kos_check...kos_check_clmul

It would be even better if we had a trait for Clmul which the different backends implement... but not totally necessary as long as we are sure that all of them provide the same impls. If that weren't the case we could get build errors.

themighty1:
> @themighty1 I think we can simplify clmul considerably. We do not need the Clmul wrapper that you wrote which does runtime checking on whether intrinsics are available. That is something that is determined at _build_ time. All backend implementations provide the same interface and a type called Clmul.
> 
> I made an example branch of how this can be done instead: [kos_check...kos_check_clmul](https://github.com/tlsnotary/tlsn/compare/kos_check...kos_check_clmul)
> 
> That branch isn't complete but hopefully you get the idea. This also avoids the test duplication that @rozbb was pointing out.
> 
> It would be even better if we had a trait for Clmul which the different backends implement... but not totally necessary as long as we are sure that all of them provide the same impls. If that weren't the case we could get build errors.

@sinui0 I see your point of simplifying by detecting intrinsics at build time.
Will this not adversely affect some future project which integrates the TLSNotary's lib and instead of shiping 1 binary, they'd have to ship 2 : with and without intrinsics support?

btw, the original Polyval crate also had that dynamic intrinsics detection.

themighty1:
> You need to add attributes to the CLMUL files

@rozbb , what specific attributes do you have in mind?

rozbb:
> @rozbb , what specific attributes do you have in mind?

Oops I meant "attributions". As in just credit RustCrypto and link to the original folder

sinui0:
> Will this not adversely affect some future project which integrates the TLSNotary's lib and instead of shiping 1 binary, they'd have to ship 2 : with and without intrinsics support?

That is convention, most projects ship platform specific binaries. The soft impl would typically be built into in the portable binary. Doing otherwise bloats the binary size and has runtime branching costs.

themighty1:
> The Polyval crate uses a Union type for `inner` and doesn't persist the `has_intrinsics` flag nor wraps `inner` in an `Option`. Can we use the same approach as that crate?

```

if self.token.get() {
            unsafe { (*self.inner.intrinsics).update(x) }
        } else {
            unsafe { (*self.inner.soft).update(x) }
        }

```

It persists a token (which server as a flag). As a byproduct of the Union,  it has to wrap calls to soft impl in unsafe{}.
Shouldn't our philosohy be that we dont sacrifice Rust's safety for the sake of extra performance, especially when performance gains are insignificant?

themighty1:
Attribution is at the top of lib.rs

themighty1:
As far as benches go: with clmul_reuse() we perform one carryless multiplication in 11ns, whereas polyval crate bench shows they do it in 1ns.

PR #39, opened 2022-07-05
WIP: The AUTHDECODE protocol

rozbb:
Added a WIP draft of the AUTHDECODE protocol to discuss here.

> This document describes a protocol that runs after a garbled-circuit two-party computation, that allows the receiver to obtain a signed commitment to their output (with the sender being the signing party). This protocol is generic over choice of polynomial commitment scheme and garbled-circuit implementation.

### DIFF
--- a/docs/authdecode.md
+++ b/docs/authdecode.md
@@ -1,0 +1,216 @@
+# The AUTHDECODE Protocol
+
+This document describes a protocol that runs after a garbled-circuit two-party computation, that allows the receiver to obtain a signed commitment to their output (with the sender being the signing party). This protocol is generic over choice of polynomial commitment scheme and garbled-circuit implementation.
+
+## Background
+
+Parties called the Notary (aka "Sender") and Requester (aka "Receiver") perform some 2-party garbled-circuit computation together. At the end, the Requester has "garbled bits" or "labels" `wᵢ ∈ 𝔽` (where 𝐅 is a finite field; see "Note on arithmetic delta" below for an explanation of why this isn't an unstructured 128-bit string). When decoded, each `wᵢ` will reveal the i-th bit of the output of the 2-party computation.
+
+## Goal
+
+The Requester needs the Notary to affirm that the resulting plaintext is the correct output of the 2PC protocol. Concretely, The Requester wants to decode `w` to plaintext `P`, and have the Notary sign a commitment to `P` _without_ the Notary learning `w` or `P`.
+
+## First steps
+
+We first describe our protocol at a high level.
+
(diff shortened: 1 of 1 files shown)


docs/authdecode.md
```diff
@@ -0,0 +1,216 @@
...
+            Let com_d = com_w - Δ·com_p
+    Prover → Verifier:
+        // Evaluate the LHS of the equation W₀ = w - Δp at c
+            Let W₀(X) = interp(W₀)
+            Let W₀c = W₀(c)
+            Let π = PolyWit(ck, W₀, (c, W₀c))
+    Verifier:
+        // Check the eval on W₀
+            Let com_W₀ = PolyCom(ck, W₀; 0)
+            Check PolyWitEval(ck, π, com_W₀, (c, W₀c))
```
xiangxiecrypto, line 133: This step seems unnecessary.

docs/authdecode.md
```diff
@@ -0,0 +1,216 @@
...
+
+Let `Wᵢⱼ' ∈ {0,1}¹²⁸` represent a bitstring label of the underlying garbled-circuit scheme. The garbler constructs its _algebraic labels_ `Wᵢⱼ ∈ 𝔽` as follows. First, it randomly selects a PRG seed `ρ`, and then generates a vector of `n` field elements `W₀ = PRG(ρ)`. The garbler also uniformly selects `Δ ∈ 𝔽`, and sets `W₁ = W₀ + (Δ, Δ, ..., Δ)`. The `W₀` and `W₁` represent the labels of the 0 bits and the 1 bits, respectively. In order to transmit the algebraic labels to the Requester, the Notary encrypts each algebraic label using the nonalgebraic label as a key. Concretely, the Notary computes `Cᵢⱼ = Enc_{Wᵢⱼ'}(Wᵢⱼ)` (using AES-GCM, since `Wᵢⱼ` is two block widths), shuffles `C` such that each `Cᵢ₀` is swapped with `Cᵢ₁` with probability 1/2, and sends `C` to the Requester.
+
+When the underlying 2PC protocol terminates, the Requester will compute `wᵢ = Enc_{wᵢ'}(Cᵢ₀)` and `wᵢ = Enc_{wᵢ'}(Cᵢ₁)`, keeping whichever decryption succeeds.
+
+Thus, we have constructed a post-GC setup whereby each `wᵢ` is a random field element, and each column of `W ∈ 𝔽²ⁿ` is related by a global, uniform Δ.
```
xiangxiecrypto, line 217: It looks like that the garbler could garble an identity gate with input bit-string labels and output arithmetic labels. Then we only need to hash  "encrypt" output labels together with the point-and-permute optimization.

docs/authdecode.md
rozbb: Yeah, so I didn't say this but this step is an optimization. The verifier could technicall compute `W₀c` itself, but we're trying to make verification cost as cheap as possible. So we let the prover do the eval for the verifier. Still though, `R_decode` requires 2 evals and it seems like overkill somehow

rozbb: Noted in latest version

docs/authdecode.md
rozbb: Cleaned up in latest version

PR #42, opened 2022-07-18
prevent GCDE leakage on the Notary side

themighty1:
This protocol prevents GCDE leakage on the Notary's side. The md file is here https://github.com/tlsnotary/tlsn/blob/gcde_fix_leak/docs/gcde_leak_fix.md

PR #46, opened 2022-07-25
DualEx with delayed check 

themighty1:


themighty1:
The doc is located here: https://github.com/tlsnotary/tlsn/blob/delayed_check.md/docs/delayed_check.md

themighty1:
Still need to decide which protocol to use when encrypting the request: this one or the one here https://github.com/tlsnotary/tlsn/pull/42

themighty1:
closing since the other approach was chosen

PR #47, opened 2022-07-26
Work on Garbled Circuits

sinui0:
List of changes:

- [x] Moved `Circuit` into mpc-circuits
- [x] Features to statically include circuits into binary
- [x] Refactored `Circuit` data model and added additional meta data
- [x] Reworked half gate implementation to be in functional form
- [x] Removed garbling from `mpc-aio` until we have a better idea of what core will look like
- [x] Move circuits from `tls-2pc-core`
- [x] Dual execution mode implementation and tests
- [x] Remove "public labels" from half gate implementation


themighty1:
ACK

sinui0:
I kept the unit tests in `tls-2pc-core` as I do think that is the right home for them. I moved the bristol-fashion circuits and assembler into it's own directory. I converted the circuits to binary format and added them as static includes into the crate. Those tests now run a lot faster because we don't have to do parsing.

sinui0:
This PR turned out to be pretty monolithic as I made sweeping changes to the garbled circuit impl. I'd like to merge this in now and build off it on dev

mpc-core/src/garble/evaluator.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn evaluate<C: BlockCipher<BlockSize = U16> + BlockEncrypt>(
+    cipher: &C,
+    circ: &Circuit,
+    input_labels: &[BinaryLabel],
+    encrypted_gates: &[EncryptedGate],
+) -> Result<Vec<BinaryLabel>, Error> {
+    let mut labels: Vec<Option<Block>> = vec![None; circ.len()];
+
+    // Insert input labels
```
themighty1, line 52: I don't like the idea of the generator setting ids on the input labels he is sending. I think he should just send them as one blob and let the evaluator assign ids.

Not only we spend extra ~3% bandwidth on the ids, but also he may do some funny business by assigning malicious ids.
(I know that right now the attack will not go through. But it may slip in later when someone decides to refactor this part of the code).

mpc-core/src/garble/circuit.rs
```diff
@@ -1,82 +1,295 @@
...
+) -> (Vec<[BinaryLabel; 2]>, Block) {
+    let delta = match delta {
+        Some(delta) => *delta,
+        None => {
+            let mut delta = Block::random(rng);
+            delta.set_lsb();
+            delta
+        }
+    };
+    let low = Block::random_vec(rng, count);
```
themighty1, line 228: let's change this to "zero_label"

mpc-core/src/garble/circuit.rs
```diff
@@ -1,82 +1,295 @@
...
+                ]
+            })
+            .collect(),
+        delta,
+    )
+}
+
+/// Returns wire labels corresponding to wire truth values
+///
+/// Panics if wire id not in label collection
```
themighty1, line 251: wire *is* not

mpc-core/src/garble/circuit.rs
```diff
@@ -1,82 +1,295 @@
...
+    if circ.input_len() != inputs.len() {
+        return Err(Error::InvalidInput(InputError::Duplicate));
+    }
+
+    Ok(inputs)
+}
+
+/// Decodes output wire labels into plaintext.
+///
+/// Using the point-and-permute (p&p) technique, the two adjacent labels
```
themighty1, line 282: I'm sorry, the wroding was confusing the first time around. Could we change this to:

// Thanks to the point-and-permute (p&p) technique, the two adjacent labels
/// will have the opposite p&p bits. We apply the decoding to the p&p bits.

mpc-core/src/garble/circuit.rs
```diff
@@ -1,82 +1,295 @@
...
+/// is encoded to 1, the other one is encoded to 0.
+pub fn decode(labels: &[BinaryLabel], decoding: &[bool]) -> Vec<bool> {
+    assert!(
+        labels.len() == decoding.len(),
+        "arrays are different length"
+    );
+    labels
+        .iter()
+        .zip(decoding)
+        .map(|(label, decode)| (label.as_ref().lsb() != 0) ^ decode)
```
themighty1, line 293: when encoding we used " == 1", can we use the same here
(label.as_ref().lsb() == 1) ^ decode)

themighty1 (changes requested):
Good work overall. Left some comments.

mpc-core/src/garble/circuit.rs
```diff
@@ -1,82 +1,146 @@
...
+/// Decodes output wire labels into plaintext
+pub fn decode(labels: &[Block], decoding: &[bool]) -> Vec<bool> {
+    assert!(
+        labels.len() == decoding.len(),
+        "arrays are different length"
+    );
+    labels
+        .iter()
+        .zip(decoding)
+        .map(|(label, decode)| (label.lsb() != 0) ^ decode)
```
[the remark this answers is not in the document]
themighty1: outdated, same q below

th4s (commented):
LGTM, nice work

mpc-circuits/src/parse.rs
```diff
@@ -1,15 +1,15 @@
...
+};
 use regex::Regex;
 use std::{
     fs::File,
     io::{BufRead, BufReader},
 };
 
 /// Parses captures into a Vec for convenience
-fn line2vec<'a>(re: &Regex, line: &'a str) -> Result<Vec<&'a str>, CircuitParserError> {
+fn line2vec<'a>(re: &Regex, line: &'a str) -> Result<Vec<&'a str>, Error> {
```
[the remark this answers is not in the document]
sinui0: I have mixed feelings on that. I prefer shorter names and to use namespacing. If I was using `foo::Error` I would just refer to it as `foo::Error`. And in cases where that is not good enough, renaming with `as` is fine.

mpc-circuits/src/circuit.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+impl From<String> for CircuitId {
+    fn from(id: String) -> Self {
+        Self(id)
+    }
+}
+
+#[derive(Clone)]
```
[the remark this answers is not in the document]
sinui0: I didn't derive Debug because I didn't want to include `gates` in it, and was moving fast. I just impl'd it.

mpc-circuits/src/circuit.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub(crate) fn new(gates: &[Gate]) -> Self {
+        let mut hasher = Sha256::new();
+        for gate in gates {
+            hasher.update(&gate.to_bytes());
+        }
+        Self(hex::encode(hasher.finalize()))
+    }
+}
+
+impl ToString for CircuitId {
```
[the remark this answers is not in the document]
sinui0: I just switched to AsRef<String> instead. I don't want to use `Display` as the functionality I was wanting was serialization, which may not be the same as `Display`.

mpc-core/src/garble/evaluator.rs
sinui0: Can you elaborate on what the issue is with denoting the ids? If someone were to refactor this code they would be responsible to not introduce vulnerabilities. I can't think of how an attack would work. The generator is always capable of lying about the labels they send. The ids just provide some structure to help with catching bugs and code flow.
`prepare_inputs` does some checks on the input labels to make sure there is no funny business.

themighty1: > If someone were to refactor this code they would be responsible to not introduce vulnerabilities.
From my experience, it is safer just not to introduce pitfalls at all.
If you don't see eye-to-eye with me, that's ok, you can mark this as resolved.

sinui0: It's not that we don't see eye-to-eye, I'm just trying to understand the issue you're seeing. If the generator is sending what appears to be uniformly random input labels to the evaluator, what can he do to trick the evaluator?

themighty1: (Nothing with the current checks.)
But potentially he can set wrong ids and overwrite evaluator's labels with his.

sinui0: We can't guard against that in `fn evaluate`, we'll have to do it sometime between deserializing the message from the generator and here. Fortunately with Rust's typesystem we can ensure this. I will add a new type `SanitizedInputLabels` which forces the user to perform these checks before evaluating the circuit.

PR #51, opened 2022-08-08
IO features mpc-circuits

sinui0:
Some more work on circuit models to make life easier at higher layers


mpc-circuits/src/circuit.rs
```diff
@@ -181,6 +392,161 @@ impl std::fmt::Debug for Circuit {
...
+    ///
+    /// Circuit wire ids are expected to be sorted and arranged in the following order:
+    ///  1. Input wires
+    ///  2. Intermediate gate wires
+    ///  3. Output wires
+    ///
+    /// All input and output wires must belong to a [`Group`].
+    ///
+    /// [`Gate`] wires can not be connected to themselves, ie the [`Circuit`] must be acyclic.
+    pub fn new(
```
th4s, line 407: Can we break this down into smaller functions?

th4s (commented):
Nice work :+1: I would just break down that one function.

themighty1 (changes requested):
lgtm, one comment.

th4s (approved):
:rocket:

PR #63, opened 2022-08-18
Oblivious Transfer Slicing API

th4s:
This pull request introduces at the `mpc-core` level the possibility for KOS15 senders and receivers to split into several instances in order to allow being consumed in different protocols.


th4s:
This has been added in #73.

PR #64, opened 2022-08-19
Improvements for padding in KOS15

th4s:
- Remove `self.padding` of KOS15Receiver
- Remove the entire padding after transposition for both sender and
  receiver

### DIFF
--- a/mpc-core/src/ot/extension/kos15/sender.rs
+++ b/mpc-core/src/ot/extension/kos15/sender.rs
@@ -306,7 +306,7 @@ where
             return Err(ExtSenderCoreError::ConsistencyCheckFailed);
         }
 
-        // Remove the last 256 rows which were sacrificed due to the KOS check
+        // Remove additional rows introduced by padding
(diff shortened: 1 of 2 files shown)


mpc-core/src/ot/extension/kos15/sender.rs
```diff
@@ -224,7 +224,7 @@ where
         //   setup
         let rem = ncols_unpadded % 256;
         let pad1 = if rem == 0 { 0 } else { 256 - rem };
```
themighty1, line 226: pad1 is not needed anymore

mpc-core/src/ot/extension/kos15/sender.rs
```diff
@@ -307,7 +307,7 @@ where
         }
 
         // Remove the last 256 rows which were sacrificed due to the KOS check
```
themighty1, line 309: the comment should be changed

themighty1 (approved):

Issue #49, opened 2022-08-01
Circuit Format & Builder


sinui0:
Right now we're using Bristol-fashion format to specify the circuits, and a [custom compiler](https://github.com/wyatt-howe/macro-circuit-assembler) for synthesizing multiple together. It would be useful to be able to build circuits using our mpc-circuits library for the following reasons:

- Circuits can be built at runtime allowing for dynamic construction and execution if needed
- A better experience prototyping circuits, testing changes and documentation would be clearer.
- Reduce binary size, as it would only have to ship subcircuits which could be assembled at runtime. For example, if a circuit contains multiple instances of a subcircuit, we only need to include that subcircuit once then replicate in memory. Similarly for if multiple circuits have common subcircuits, eg. our application circuits contain multiple instances of AES, SHA, adder64, etc.

We should move away from the bristol-fashion format in general, as it is lacking in many ways, eg. file size, no circuit metadata, heavy to parse.

PR #72, opened 2022-09-06
Circuit Builder API

sinui0:
This PR includes an implementation of CircuitBuilder which can be used to synthesize custom circuits. See https://github.com/tlsnotary/tlsn/issues/49

Also included:

- Two constant wire types to simplify circuits
- A binary for converting bristol-fashion circuits to a yaml format (`CircuitSpec`). This is used to add additional metadata to circuits (naming, version, value types)

### DIFF
--- a/mpc-circuits/src/builder.rs
+++ b/mpc-circuits/src/builder.rs
@@ -84,7 +84,7 @@ impl CircuitHandle {
         self.inputs.get(id).cloned()
     }
 
-    /// Returns a handle ot the sub-circuit output
+    /// Returns a handle to the sub-circuit output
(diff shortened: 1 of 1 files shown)


mpc-circuits/src/builder.rs
```diff
@@ -0,0 +1,694 @@
...
+///
+/// It has three states:
+/// 1. Define inputs
+/// 2. Define gates
+/// 3. Define outputs
+#[derive(Debug)]
+pub struct CircuitBuilder<S: BuilderState>(S);
+
+impl CircuitBuilder<Inputs> {
+    /// Creats new builder
```
themighty1, line 270: Creates

themighty1 (approved):
good work!

PR #76, opened 2022-09-17
Refactor 2pc-handshake to typestates

sinui0:
This PR refactors our 2pc handshake implementation to use the typestate pattern and renames from Master/Slave to Leader/Follower.

Business logic remains untouched, as can be seen by the minimal changes to the `test_prf()` unit test.

### DIFF
--- a/tls-2pc-core/src/handshake/mod.rs
+++ b/tls-2pc-core/src/handshake/mod.rs
@@ -4,8 +4,8 @@ pub mod sha;
 mod utils;
 
 pub use crate::msgs::handshake::HandshakeMessage;
-pub use follower::HandshakeFollower;
-pub use leader::HandshakeLeader;
+pub use follower::{state as follower_state, HandshakeFollower};
+pub use leader::{state as leader_state, HandshakeLeader};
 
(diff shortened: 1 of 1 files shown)


th4s (approved):
nice, ACK

themighty1 (approved):
Looks good but why were the comments not preserved @sinui0 ?

themighty1:
> @themighty1 Oh good catch, are you referring to the comments on the traits? I overlooked those. They don't quite map to the new code as we won't have traits in core for this. However, I'll add back what I can and more documentation in #78

I actually meant comments like this one https://github.com/tlsnotary/tlsn/blob/d8b9f05a92874665a81285a9a4320b37f27f15fa/tls-2pc-core/src/handshake/master.rs#L65

I realize that they don't map 1:1 to the new code but at least some comments are needed next to the functions. Specifically it should be mentioned that Ms stands for master secret and Ke for Key expansion, Cf/Sf for ... etc.

themighty1:
Ah, i was reviewing PR one by one, I just noticed that PR78 subsumes this PR, so nvm.

PR #77, opened 2022-09-18
Refactor point addition to typestates

sinui0:
This PR refactors our point addition protocol to use typestates and renames from Master/Slave to Leader/Follower.

Business logic remains unchanged.

### DIFF
--- a/mpc-core/src/point_addition/mod.rs
+++ b/mpc-core/src/point_addition/mod.rs
@@ -44,8 +44,8 @@ impl Add<P256SecretShare> for P256SecretShare {
 impl TryFrom<BigInt> for P256SecretShare {
     type Error = BigInt;
 
-    fn try_from(key: BigInt) -> Result<Self, BigInt> {
-        Ok(Self(key.to_bytes_array::<32>().ok_or(key)?))
+    fn try_from(share: BigInt) -> Result<Self, BigInt> {
+        Ok(Self(share.to_bytes_array::<32>().ok_or(share)?))
     }
 }
 


th4s (approved):
nice, ack

mpc-core/src/point_addition/mod.rs
```diff
@@ -5,50 +5,51 @@
...
-pub use master::PointAdditionMaster;
-pub use slave::PointAdditionSlave;
+pub use follower::{state as follower_state, PointAdditionFollower};
+pub use leader::{state as leader_state, PointAdditionLeader};
 
 /// NIST P-256 Prime
 pub const P: &str = "ffffffff00000001000000000000000000000000ffffffffffffffffffffffff";
 
-pub type SecretShare = BigInt;
+/// Additive secret share of a NIST P-256 private key
```
themighty1, line 21: This should read:
/// Additive secret share of the resulting X coordinate.

And also below it should be referred to as "share" rather than "key"

mpc-core/src/point_addition/follower.rs
```diff
@@ -0,0 +1,279 @@
+//! 2-Party Elliptic curve secret-sharing using Paillier Cryptosystem
```
themighty1, line 1: This comment is vague, plus mod.rs already explain what's going on.

mpc-core/src/point_addition/mod.rs
```diff
@@ -5,50 +5,51 @@
 //! would also be possible using this approach, but it hasn't been
 //! implemented here).
```
themighty1, line 6: Can we also leave a link here to the docs where this protocol can be referenced?

themighty1 (approved):
ACK + comments

PR #78, opened 2022-09-19
PRF rename and docs

sinui0:
This PR renames our handshake protocol _back_ to PRF. In the name of modularity the PRF will be a smaller component of the handshake protocol.

I've added some documentation to the PRF module explaining what is happening and also renamed some state variables. I also added some state accessors so the aio wrapper can get at some of the state needed to pass to garbled circuits.

I also noticed that the Follower was passing p2 back to the Leader during master secret calculation. This is unnecessary as the Follower (Notary) can provide that as an input to the garbled circuit. So I removed that message from being sent.

### DIFF
--- a/tls-2pc-core/src/prf/mod.rs
+++ b/tls-2pc-core/src/prf/mod.rs
@@ -1,6 +1,8 @@
 //! This module contains the protocol for computing TLS SHA-256 HMAC PRF using 2PC in such a way
 //! that neither party learns the session keys, rather they learn respective XOR shares of the keys.
 //!
+//! For a more comprehensive explanation of this protocol see our [documentation](https://tlsnotary.github.io/docs-mdbook)
+//!
 //! To save some compute and bandwidth, the PRF can be broken down into smaller units where some can be
 //! computed without using 2PC.
 //!


th4s (approved):
Awesome documentation :rocket:

tls-2pc-core/src/prf/mod.rs
```diff
@@ -1,11 +1,82 @@
+//! This module contains the protocol for computing TLS SHA-256 HMAC PRF using 2PC in such a way
```
themighty1, line 1: Can we add a step-by-step description or a link thereto similar to https://github.com/tlsnotary/circuits/blob/9f6864934b2be9a5745c96fb40f40dfe3b6f690f/README#L35
?

themighty1:
Do you have a specific reason why the final state is not Complete? As it is now, the final state can be called multiple times.

themighty1 (approved):
thanks, utACK

themighty1:
> Can you elaborate on this? The protocol is terminating on both Leader and Follower, as it consumes self

When `follower` is in state Sf2 and the code calls
`let follower_msg = follower.next(leader_msg);`
after that the variable `follower` is still in the state Sf2 and can be mistakenly called again with `follower.next(...)`. Whereas if it returned `follower` in the state `Complete` like
`let (follower_msg, follower) = follower.next(leader_msg);`
then such programmer's error would be ruled out.

themighty1:
ok, I see now.

PR #83, opened 2022-09-21
OT mpc-aio + utils

sinui0:
This PR includes the implementation for extended OT in the aio layer, as well as some utilities such as the stream muxer (which is WIP).

We're not shooting for perfect here, some aspects may change but for now I'd like to get this stuff merged to avoid monolithic PRs as we start compounding on this crate.

### DIFF
--- a/utils-aio/Cargo.toml
+++ b/utils-aio/Cargo.toml
@@ -7,25 +7,26 @@ edition = "2021"
 name = "utils_aio"
 
 [features]
-default = ["muxer"]
+default = []
 codec = []
 muxer = ["dep:tokio"]
+duplex = []
 
 [dependencies]
-bytes = "1.1.0"
+bytes = "1.1"
 prost = "0.9"
-tokio = { version = "1.21.0", optional = true }
-tokio-util = { version = "0.7.1", features = ["codec", "compat"] }
+tokio = { version = "1.21", optional = true }
+tokio-util = { version = "0.7", features = ["codec", "compat"] }
 async-tungstenite = { version = "0.16" }
 futures = "0.3"
 futures-util = "0.3"
-async-trait = "0.1.57"
-yamux = "0.10.2"
-thiserror = "1.0.34"
-async-std = "1.12.0"
+async-trait = "0.1"
+yamux = "0.10"
+thiserror = "1.0"
+async-std = "1.12"
 
 [dev-dependencies]
-tokio = { version = "1.21.0", features = [
+tokio = { version = "1.21", features = [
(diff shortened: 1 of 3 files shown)


themighty1 (approved):
ACK

mpc-aio/src/protocol/ot/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+#[async_trait]
+pub trait ObliviousReceive {
+    type Choice;
+    type Outputs;
+
+    async fn receive(&mut self, choices: &[Self::Choice]) -> Result<Self::Outputs, OTError>;
+}
+
+#[cfg(test)]
+mockall::mock! {
```
th4s, line 46: Is that used anywhere?

utils-aio/Cargo.toml
```diff
@@ -6,12 +6,28 @@ edition = "2021"
...
+[features]
+default = ["muxer"]
+codec = []
+muxer = ["dep:tokio"]
 
 [dependencies]
 bytes = "1.1.0"
 prost = "0.9"
-tokio-util = { version = "0.7.1", features = ["codec"] }
+tokio = { version = "1.21.0", optional = true }
```
th4s, line 17: Is there a good reason for specifying dependencies that tightly. My understanding was that as a general rule of thumb:
- If there is a X.x.x version, where X != 0, then  use "X" as a dependency
- If X == 0, then use "0.x" as a dependency

Otherwise we opt out of potentially useful non-breaking updates and it could make the build process more complicated when having to deal with several very similar versions of the same crate.

utils-aio/src/lib.rs
```diff
@@ -1 +1,5 @@
 pub mod codec;
+pub mod duplex;
```
th4s, line 2: This module is only used in tests so far. Add `#[cfg(test)]`

utils-aio/src/muxer.rs
```diff
@@ -0,0 +1,318 @@
...
+    pub fn control(&self) -> YamuxMuxerControl {
+        YamuxMuxerControl {
+            control_sender: self.control_sender.clone(),
+        }
+    }
+
+    /// Processes pending substreams
+    ///
+    /// This function must be called repeatedly to make progress
+    pub async fn next_stream(&mut self) -> Result<(), MuxerError> {
```
th4s, line 121: If we have time, break down into digestible pieces.

th4s (changes requested):
Nice work :rocket:. Some small nitpicks.

PR #87, opened 2022-09-22
Misc fixes

themighty1:
I will collect here misc fixes which can be merged later after the bigger PRs are merged


themighty1:
pushed to dev directly

PR #88, opened 2022-09-25
Send constant wires to evaluator

sinui0:
This PR fixes a bug where the wire labels for constant inputs are not sent to the evaluator. The fix is quite simple. We now filter the circuit inputs for constants and include their labels with the garbled circuit data sent to the evaluator.


th4s (approved):
ACK

themighty1 (approved):
ACK

PR #95, opened 2022-10-04
Introduce macro for extracting enum variants in messaging

th4s:


th4s:
Not really a difference, but when using braces rust-fmt puts it into a single line. When using parentheses it uses 5 lines. I also prefer parentheses, but thought that it is nicer when it uses only a single line. What do you think?

PR #99, opened 2022-10-06
Async compute for garbled circuits

sinui0:
This PR introduces `Generator` and `Evaluator` traits into mpc-aio so we can move and await heavy compute on another thread seperately from the execution protocols if needed.


themighty1 (approved):
nice, pushed misc comments

th4s (approved):
Nice, good job!

PR #102, opened 2022-10-07
Use new feature workspace dependencies

th4s:
This PR introduces inheriting workspace dependencies.

It also
- bumps all sub-crate rust versions to 2021
- relax version dependencies

The sub-crate tls-client has been excluded from using workspace dependencies, because it is currently excluded from the workspace.


themighty1 (approved):
nice, tested, ACK

PR #105, opened 2022-10-13
Authdecode

themighty1:
Implemented the AuthDecode protocol, see comments in lib.rs
The aio part will follow later once we agree how this protocol will interface with the committed GC protocol.
The diagram for the protocol and the halo2 circuit is included.
Some tests are ignored since they are expensive.
Benchmarks show that it takes 1 sec on 1 core to convert 1 KB of GC output into a Poseidon hash commitment.


maceip:
✒️🔗

🆒

PR #108, opened 2022-10-27
A2M 2PC

th4s:
This PR implements an addition-to-multiplication 2pc algorithm. It depends on #106, which needs to be merged first.


gf2-128/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///   * `MaskedPartialValue` - Used for oblivious transfer
+    pub fn to_multiplicative(&self) -> (MulShare, MaskedPartialValue) {
+        let mut rng = ChaCha12Rng::from_entropy();
+
+        let random: u128 = rng.gen();
+        let mut masks: [u128; 128] = std::array::from_fn(|_| rng.gen());
+        masks[127] = masks.into_iter().take(127).fold(0, |acc, i| acc ^ i);
+
+        let mul_share = MulShare::new(inverse(random));
+        let b0: [u128; 128] =
```
themighty1, line 95: I find that I have to pause and to think what `&` does here. I think should put a comment somewhere along the lines of 
// self.inner() & (1 << i) extracts self.inner()'s bit in position `i` (counting from the right) shifted left by `i`

gf2-128/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    /// Turn into a multiplicative share and masked partial values
+    ///
+    /// This function returns
+    ///   * `MulShare` - The sender's multiplicative share
+    ///   * `MaskedPartialValue` - Used for oblivious transfer
+    pub fn to_multiplicative(&self) -> (MulShare, MaskedPartialValue) {
+        let mut rng = ChaCha12Rng::from_entropy();
+
+        let random: u128 = rng.gen();
```
themighty1, line 89: Let's make sure that random != 0 , otherwise the security will be broken

themighty1 (approved):
lgtm

PR #109, opened 2022-10-28
Semi-honest GC

sinui0:
This PR implements semi-honest garbling execution mode in `mpc-core` and a couple other minor tweaks.

We need this execution mode for constructing our "one-sided" protocol. I didn't implement the wrapper in mpc-aio as I'm still trying to sort out what the traits will look like. I've found it quite challenging to come up with good trait definitions which will support our more advanced requirements.


th4s (approved):
lgtm :rocket:, one small nitpick

themighty1 (changes requested):
ack, when feedback is addressed pls ping me for re-review, i want to push some code comments

PR #115, opened 2022-11-11
Rayon Garbler

sinui0:
This PR implements Async Garbling/Evaluation using Rayons threadpool.

### DIFF
--- a/mpc-aio/src/protocol/garble/backend/rayon.rs
+++ b/mpc-aio/src/protocol/garble/backend/rayon.rs
@@ -32,7 +32,7 @@ impl Generator for RayonBackend {
         });
         receiver
             .await
-            .map_err(|_| GCError::GarblerError("channel error".to_string()))?
+            .map_err(|_| GCError::BackendError("channel error".to_string()))?
     }
 }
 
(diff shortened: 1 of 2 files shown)


mpc-aio/src/protocol/garble/exec/dual.rs
```diff
@@ -102,7 +91,7 @@ where
         let gc_ev = GarbledCircuit::<Partial>::from_msg(circ, msg)?;
         let labels_ev = self.label_receiver.receive_labels(inputs.to_vec()).await?;
 
-        let evaluated_gc = self.evaluator.evaluate(gc_ev, &labels_ev).await?;
+        let evaluated_gc = self.garbler.evaluate(gc_ev, labels_ev).await?;
```
themighty1, line 94: can this be called `dualex_party` instead of `garbler`? because `garbler.evaluate` doesn't sound right
Or better still, can we have `self ` methods `generate` and `evaluate` which internally call garbler/evaluator's generate/evaluate?

themighty1 (commented):
shouldn't there be `pub mod garbler;` in protocol/garble/mod.rs ?
when I added it, I saw a few rust errors in rayon.rs

th4s (changes requested):
Same as Dan, you forgot to include `mod garbler` in `.../garble/mod.rs`. When I include it there are some lifetime errors.

th4s (approved):
ACK

PR #116, opened 2022-11-14
Multiplex Actors

sinui0:
This PR implements a stream multiplexer using the xtra actor framework and yamux. Additionally, it defines some traits to abstract the underlying multiplexing protocol. Finally, I slapped together a mock implementation for multiplexing which uses channels, which helps us test without having to handle serialization.

### DIFF
--- a/actor-mux/src/yamux.rs
+++ b/actor-mux/src/yamux.rs
@@ -197,8 +197,7 @@ impl YamuxMuxControl {
 
 #[async_trait]
 impl MuxControl for YamuxMuxControl {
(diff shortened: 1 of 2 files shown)


th4s (commented):
ACK, with some comments

Most of my comments I have done before knowing that `mock.rs` is basically just test code, so you can ignore most of that part.

themighty1:
will review shortly

themighty1 (approved):
ACK. added some extra code comments. Also we'll need to address the points which I raised on Discord.

PR #117, opened 2022-11-15
OT: Add sender input for the number of OTs to generate

th4s:
This PR adds the feature that for KOS OT, the sender now needs to input the number of OTs, he wants to generate. If this does not meet the receiver's value in the setup message, an error is returned.


themighty1 (approved):
ack

PR #126, opened 2022-12-09
GC Refactor + Docs

sinui0:
This PR tidies up GC a bit, moving states into a dedicated module and adds a new summary state. It also renames "label encodings" as requested by @themighty1 . Finally, I went around and added a bit more documentation.


themighty1 (approved):
ack with some comment fixes

PR #127, opened 2022-12-10
Consolidate utils

sinui0:
This PR creates a new crate which consolidates some of the utility functions we use. I also implemented some helper traits for duplicate detection.


themighty1 (approved):
ACK

PR #128, opened 2022-12-11
GC Unchecked types

sinui0:
This PR decouples the GC message layer from data validation. I've introduced "unchecked" types which are structurally correct but have not been validated against a circuit specification.

Note that I did not invest much into optimizing validation, and we still need more comprehensive tests. I'm punting both to a future PR.

This PR depends on #127 so the diff will be a bit cluttered until we get that merged and I rebase this.

### DIFF
--- a/mpc-core/src/garble/circuit.rs
+++ b/mpc-core/src/garble/circuit.rs
@@ -823,8 +823,8 @@ pub(crate) mod unchecked {
         ) -> Result<Vec<OutputValue>, Error> {
             // Validate circuit id
             if &self.circ_id != circ.id() {
-                return Err(Error::PeerError(format!(
-                    "Received evaluated circuit with wrong id: expected {}, received {}",
+                return Err(Error::ValidationError(format!(
+                    "Received garbled output with wrong id: expected {}, received {}",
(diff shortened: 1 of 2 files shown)


mpc-core/src/garble/exec/semihonest.rs
```diff
@@ -220,44 +228,19 @@ mod tests {
...
 
         let follower_labels = input_labels[1].select(&follower_input).unwrap();
-        let error = follower.evaluate(msg, &[follower_labels]).unwrap_err();
+        let error = follower
+            .evaluate(gc_partial.into(), &[follower_labels])
+            .unwrap_err();
 
         assert!(matches!(error, Error::InvalidOutputLabelCommitment));
     }
-
```
themighty1, line 236: Why was this removed?

themighty1: I can't find where. Could you point to that place?

themighty1: I see what you mean. Then this test can be considered a test of that check.  A test can always prevent a future refactor error, so I don't see a reason to remove it.

PR #131, opened 2022-12-14
Maliciously secure M2A/A2M

th4s:
This PR implements recording and replay functionality for the conversion algorithms M2A/A2M over GF(2^128).

### DIFF
--- a/share-conversion-aio/src/gf2_128/mod.rs
+++ b/share-conversion-aio/src/gf2_128/mod.rs
@@ -130,10 +130,10 @@ mod tests {
         });
 
         let (sender_output, receiver_output) = tokio::join!(sender_task, receiver_task);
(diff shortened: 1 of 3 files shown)


themighty1 (changes requested):
Pushed misc comments.
We'll need to resolve the replay issue raised on discord.

it won't let me comment in the review on this line:

https://github.com/tlsnotary/tlsn/blob/25c1d656c7a9973b54da6e80ad2feb324a4edb88/share-conversion-core/src/gf2_128/conversion/mod.rs#L46

Can we rename this to "from_sender_values(values: &[u128])" because otherwise it conveys as  `value` has the choice bits. 


Please ping me again for a review.

themighty1 (approved):
ACK to merge into dev.

PR #133, opened 2022-12-15
GC Opening

sinui0:
This PR implements the opening of a garbled circuit, and refactors circuit validation a bit using the introduced data structure. I've also implemented the msg layer types.

I'll note that this is not the most bandwidth efficient way to implement this, as parallel openings will send $\Delta$ multiple times. Additionally, sending the input label decoding for labels which the Evaluator already knows the plaintext value is redundant.

For the sake of momentum I'm punting those optimizations for later.

Lastly, I'm temporarily hardcoding the AES cipher for garbling. This will be addressed soon, I have a PR in mind which will pull this out into `GarbledCircuitConfig`.


themighty1 (commented):
pushed some comments. let's resolve my q first, then i will ack.

themighty1:
ACK

PR #136, opened 2022-12-20
Actor share conversion

th4s:
This PR implements the actor layer for share conversion. It is still incomplete and does not compile.

@sinui0 I would be interested in your opinion. The `Handle` trait only uses `&mut self`, but `SendTape` requires `self`. I see some solutions, but non which I thought of as a good ones. The  compiler errors will directly point you to what is the issue here.

### DIFF
--- a/actor-share-conversion/src/gf2_128/receiver.rs
+++ b/actor-share-conversion/src/gf2_128/receiver.rs
@@ -1,4 +1,4 @@
-use super::{A2MMessage, M2AMessage, VerifyTapeMessage};
+use super::{A2MMessage, M2AMessage, Setup, VerifyTapeMessage};
 use mpc_aio::protocol::ot::{OTFactoryError, ObliviousReceive};
 use mpc_core::{ot::config::OTReceiverConfig, Block};
 use share_conversion_aio::{
@@ -63,33 +63,6 @@ where
     }
 }
 
-impl<T, OT, U, V, W> Receiver<T, OT, U, V, W>
-where
-    T: AsyncFactory<OT, Config = OTReceiverConfig, Error = OTFactoryError> + Send,
-    OT: ObliviousReceive<bool, Block>,
-    U: Gf2_128ShareConvert,
-    V: MuxChannelControl<Gf2ConversionMessage>,
-    W: Recorder<U>,
-{
-    pub async fn setup(&mut self) -> Result<(), ShareConversionError> {
-        // We need to own the state, so we use this only as a temporary modification
-        let state = std::mem::replace(&mut self.state, State::Error);
-
-        let State::Initialized {id, mut muxer, receiver_factory} = state else {
-            return Err(ShareConversionError::Other(String::from("Actor has to be in the Initialized state")));
-        };
-
-        let channel = muxer
-            .get_channel(id.clone())
-            .await
-            .map_err(|err| ShareConversionError::Other(err.to_string()))?;
-        let receiver = IOReceiver::new(receiver_factory, id, channel);
-        self.state = State::Setup(receiver);
-
-        Ok(())
-    }
-}
-
 /// The controller to talk to the local conversion receiver actor. This is the only way to talk
 /// to the actor.
 pub struct ReceiverControl<T>(Address<T>);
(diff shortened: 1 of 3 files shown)


actor-share-conversion/src/gf2_128/sender.rs
```diff
@@ -0,0 +1,229 @@
...
+where
+    T: AsyncFactory<OT, Config = OTSenderConfig, Error = OTFactoryError> + Send,
+    OT: ObliviousSend<[Block; 2]> + Send,
+    U: Gf2_128ShareConvert + Send,
+    V: MuxChannelControl<Gf2ConversionMessage>,
+    W: Recorder<U>,
+{
+    pub async fn setup(&mut self) -> Result<(), ShareConversionError> {
+        // We need to own the state, so we use this only as a temporary modification
+        let state = std::mem::replace(&mut self.state, State::Complete);
```
themighty1, line 74: shouldn't this be explicitely `State::Error` instead of`State::Complete`? Otherwise this causes the code reviewer to pause and wonder if there may be security implications.

themighty1: yes. ::Temporary is fine

actor-share-conversion/src/gf2_128/sender.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        let State::Initialized {id, barrier, mut muxer, sender_factory} = state else {
+            self.state = state;
+            return Err(ShareConversionError::Other(String::from("Actor has to be initialized")));
```
themighty1, line 81: I don't understand the meaning of this error message. As I understand ::Initialized is the original state and it is not even possible to call .setup() without first going through the ::Initialized state. Maybe this msg should be "wrong state"?

actor-share-conversion/src/gf2_128/sender.rs
```diff
@@ -0,0 +1,229 @@
...
+    V: MuxChannelControl<Gf2ConversionMessage> + Send + 'static,
+    U: Gf2_128ShareConvert + Send + 'static,
+    IOSender<T, OT, U, Tape>: SendTape,
+{
+    type Return = Result<(), ShareConversionError>;
+
+    async fn handle(&mut self, _message: SendTapeMessage, ctx: &mut Context<Self>) -> Self::Return {
+        let state = std::mem::replace(&mut self.state, State::Complete);
+        let State::Setup(state) = state else {
+            return Err(ShareConversionError::Other(String::from(
```
themighty1, line 232: here, unlike in setup() self is not returned into the previous state. i think it should be reverted to the previous state if this is not a critical error.

actor-share-conversion/src/gf2_128/receiver.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    async fn handle(
+        &mut self,
+        _message: VerifyTapeMessage,
+        ctx: &mut Context<Self>,
+    ) -> Self::Return {
+        let state = std::mem::replace(&mut self.state, State::Complete);
+
+        let State::Setup(state) = state else {
+            return Err(ShareConversionError::Other(String::from(
```
themighty1, line 235: the state should be restored just like my comment in sender.rs

themighty1 (changes requested):
Well done. pushed some comments
Let's resolve my Qs before I ACK

actor-share-conversion/src/gf2_128/sender.rs
themighty1: ok, then let's word the error "Actor has to be in the Initialized state" to remove ambiguity

actor-share-conversion/src/gf2_128/sender.rs
themighty1: sry, i meant that it should be reverted just before the line
`return Err(ShareConversionError::Other(String::from(
                "Actor is not setup"
            )));
`
not after `state.send_tape().await`

If it was up to me, I would make it so that all errors when the handler is called in the wrong state are critical errors and would shut down the Actor. 
But since you chose a less paranoid approach, we should revert the state when a non-critical error happens.

actor-share-conversion/src/gf2_128/receiver.rs
themighty1: same as clarification above, the reverting should happen before the line
`return Err(ShareConversionError::Other(String::from(
                "Actor is not setup"
            )));`

themighty1 (approved):
ack

PR #137, opened 2022-12-22
Refactor OT traits to be generics

sinui0:
This PR refactors OT traits to use generics instead of associated types. It also changes them to take inputs by value instead of reference.

### By value instead of reference
I realized that any implementation of these traits will end up cloning/copying the inputs anyhow. So we may as well pass by value so that the futures are `Send`.

### Generics instead of associated types
It makes more sense to have these traits be generic so we can provide clean blanket impls as I've demonstrated in this PR. With associated types a struct is restricted to implementing the trait once and forces us to create new traits.

### DIFF
--- a/mpc-aio/src/protocol/ot/mock.rs
+++ b/mpc-aio/src/protocol/ot/mock.rs
@@ -1,7 +1,8 @@
 use std::sync::{Arc, Mutex};
 
 use super::{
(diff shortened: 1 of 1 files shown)


themighty1:
lmk when this is ready for review

themighty1 (approved):
ack

PR #141, opened 2022-12-31
Refactor Wire Groups

sinui0:
This PR refactors our model for circuit wire groups. I introduced a trait called `WireGroup` which provides a better abstraction for a group of wires. Following that, the `Input` and `Output` types are now simple newtypes which wrap a `Group` using an `Arc` reference so they are essentially free to clone. Similarly I introduced the `GroupValue<G>` type which helps out with dealing with values.

This provides a better abstraction to work with. I was encountering an awful amount of compounding code duplication in `mpc-core` without having this.

### DIFF
--- a/mpc-circuits/src/circuit.rs
+++ b/mpc-circuits/src/circuit.rs
@@ -255,18 +255,18 @@ pub struct Circuit {
     /// Total number of XOR gates
     pub(crate) xor_count: usize,
 
-    /// All input ids in ascending order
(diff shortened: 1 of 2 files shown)


themighty1 (approved):
ACK

I added a check on const inputs. Feel free to move this check to a different place, maybe to Input::new or Group::new if you feel it more logical to check it elsewhere.

I'll open an issue to create tests for the validation errors once this is merged.

PR #142, opened 2022-12-31
Refactor Wire Labels

sinui0:
This PR builds on top of #141 to introduce a better abstraction for wire labels. It's rather large, sorry. A lot of the changes are simply renaming types across the crate.

Changes:
- Abstract labels to be generic over a `WireGroup` and `Full` vs `Active` states
- Introduce type definitions for `(Full/Active)(Input/Output)Labels` for clarity
- Tidy up some types such as `OutputCheck` -> `LabelsDigest`
- `GarbledCircuit<Full>` and `GarbledCircuit<Evaluated>` no longer holds on to all the wire labels, only the Input and Output labels
- Consolidated many error types to `LabelError`

### DIFF
--- a/mpc-core/src/garble/circuit.rs
+++ b/mpc-core/src/garble/circuit.rs
@@ -73,8 +73,11 @@ pub mod state {
     /// Full garbled circuit data. This includes all wire label pairs, encrypted gates and delta.
     #[derive(Debug)]
     pub struct Full {
+        /// Full input labels sorted ascending by id
         pub(crate) input_labels: Vec<FullInputLabels>,
+        /// Full output labels sorted ascending by id
         pub(crate) output_labels: Vec<FullOutputLabels>,
+        /// Encrypted gates sorted ascending by id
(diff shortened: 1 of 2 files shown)


mpc-core/src/garble/label/state.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.low
+            .iter()
+            .map(|low| low.to_pair(self.delta, false))
+            .collect()
+    }
+
+    pub(super) fn from_decoding(
+        active: Active,
+        delta: Delta,
+        decoding: Vec<bool>,
```
themighty1, line 55: why is `decoding` not `Vec<LabelsDecodingInfo>`? It would be much cleaner, imo
It is only incidental that the decoding is a simple XOR operation. What if it was a complex function? We wouldn't want to copy-paste this function around.



This will require modifying Labels::decode() into Labels::decode_to_bits() and Labels::decode_to_group_value

themighty1 (approved):
ACK. left some qs.

PR #143, opened 2023-01-02
Arc Labels + Iterators

sinui0:
This PR builds on top of #142.

It wraps the underlying label data in `Arc` to make it cheap to clone, and thus more memory efficient when re-using labels between garbled circuit executions. On that note, I added a method for converting from output labels to input labels.

I also replaced some of the accessors to return iterators instead

### DIFF
--- a/mpc-core/src/garble/label/mod.rs
+++ b/mpc-core/src/garble/label/mod.rs
@@ -159,7 +159,7 @@ where
     /// Validates whether the provided active labels are authentic
     pub fn validate(&self, labels: &Labels<G, state::Active>) -> Result<(), LabelError> {
         for (pair, label) in self.state.iter().zip(labels.iter()) {
-            if label.value != pair.low() && label.value != pair.high() {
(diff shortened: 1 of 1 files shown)


mpc-core/src/garble/label/mod.rs
```diff
@@ -146,10 +158,8 @@ where
 
     /// Validates whether the provided active labels are authentic
     pub fn validate(&self, labels: &Labels<G, state::Active>) -> Result<(), LabelError> {
-        for (pair, label) in self.state.to_labels().into_iter().zip(&labels.state.labels) {
-            if label.value == pair.low() || label.value == pair.high() {
-                continue;
-            } else {
+        for (pair, label) in self.state.iter().zip(labels.iter()) {
+            if label.value != pair.low() && label.value != pair.high() {
```
themighty1, line 162: This is harder to understand as opposed to a simple
` if ! (label.value == pair.low() || label.value == pair.high()) `

themighty1 (approved):
ACK with a comment

PR #146, opened 2023-01-03
Async Factory

sinui0:
This PR builds on top of #137. Instead of having a different trait for every kind of factory, I've introduced a generic trait `AsyncFactory`.

I've also introduced config types for OT to accomodate this change.

This helps provide more flexibility in trait bounds. For example, right now OTSenderFactory looks like this:

```rust
#[async_trait]
pub trait OTSenderFactory<T> {
    type Protocol: ObliviousSend<T> + Send;

    /// Constructs a new Sender
    ///
    /// * `id` - Instance id
    /// * `count` - Number of OTs to provision
    async fn new_sender(
        &mut self,
        id: String,
        count: usize,
    ) -> Result<Self::Protocol, OTFactoryError>;
}
```

but the problem is we have other trait bounds we may want to put on the item returned from this factory, eg `ObliviousReveal`. In the current pattern we would need a new trait for every combination of trait bound.

### DIFF
--- a/actor-ot/src/lib.rs
+++ b/actor-ot/src/lib.rs
@@ -186,12 +186,12 @@ mod test {
             .collect();
 
         let mut sender = sender_control
-            .new(instance_id.clone(), sender_config(choices.len()))
+            .create(instance_id.clone(), sender_config(choices.len()))
             .await
             .unwrap();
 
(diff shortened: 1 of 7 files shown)


themighty1 (approved):
ack

PR #148, opened 2023-01-04
Label Encoder

sinui0:
This PR builds on #142.

It implements an encoder which can be used for input label generation in garble circuits. With this we can generate all labels with 1 seed and conveniently partition them into sets as needed using the domain + stream id functionality (1 nonce per set).

For example, for encryption/decryption we can use this to nicely group the plaintext labels together. By convention we can say that the plaintext labels are always generated from stream 0. The tx labels will be domain 0, rx labels domain 1. Any other labels we don't need to use past garbling can be pulled from another stream id.

I also was interested to see encoding performance, so for those curious it takes ~3 ${\mu}s$ to generate input labels for the AES circuit on my machine.

### DIFF
--- a/mpc-core/src/garble/label/encoder.rs
+++ b/mpc-core/src/garble/label/encoder.rs
@@ -107,9 +107,9 @@ mod test {
     fn test_encoder(circ: Circuit) {
         let mut enc = ChaChaEncoder::new([0u8; 32], 0);
 
-        circ.inputs()
-            .iter()
-            .for_each(|input| _ = enc.encode(input.id() as u32, input))
+        for input in circ.inputs() {
+            enc.encode(input.id() as u32, input);
+        }
     }
 
     #[rstest]


mpc-core/src/garble/label/encoder.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Sets the selected stream id, restoring word position if a stream
+    /// has been used before.
+    fn set_stream(&mut self, id: u32) {
+        //           MSB -> LSB
+        //   31 bits   32 bits   1 bit
+        //   [domain]   [id]   [reserved]
+        // The reserved bit ensures that we never pull from stream 0 which
+        // is reserved to generate delta
+        let new_id = ((self.domain as u64) << 33) + ((id as u64) << 1) + 1;
```
themighty1, line 70: What if this overflows and new_id becomes 0 ?

mpc-core/src/garble/label/encoder.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        let current_id = self.rng.get_stream();
+
+        // noop if stream already set
+        if new_id == current_id {
+            return;
+        }
+
+        // Store word position for current stream
```
themighty1, line 79: This will not store a new value if an old value is already present

themighty1:
@sinui0 , Im not sure which files to look at rn. Seems like multiple PRs got merged.

themighty1 (approved):
ACK
I had to change the code of the test since the way it was written was tripping up my rust-analyzer

PR #149, opened 2023-01-09
Ghash core implementation

th4s:
This PR implements the ghash core module. This is an older PR, which to some degree has already been reviewed, but did not make it into dev, because we first focused on the share-conversion crates. I have adapted it to the current state of `dev` branch.


tls-2pc-core/src/ghash/mod.rs
```diff
@@ -1,295 +1,395 @@
...
+pub struct Intermediate {
+    odd_mul_shares: Vec<u128>,
+    cached_add_shares: Vec<u128>,
+}
 
-    /// Returns true when the protocol is complete.
-    fn is_complete(&mut self) -> bool;
+/// Final state for Ghash protocol
+///
+/// This is when each party can compute a final share of the MAC, because both now have
```
themighty1, line 45: Can we move away from saying "MAC" to saying "Ghash output" or similar.

themighty1 (changes requested):
i'll need to have my Qs answered to better understand the code and then i'll review again.

themighty1:
I had to simplify the function in order to wrap my head around it. lmk if you don't think it is simpler that way.

themighty1 (approved):
ACK

PR #152, opened 2023-01-11
Circuit Cyclic Reference

sinui0:
This PR adds `Weak<Circuit>` references to wire groups so that the circuit a group is from can always be retrieved. I also did some tidying up of errors, and added an `UncheckedGroup` type which helps accomodate this feature.

@themighty1 I relocated the "const value must have 1 wire check" to `ValueType::valid_length`

Apologies in advance for how many files this touches. I've separated changes into 2 commits. The second commit is just updating all the various dependents.


themighty1 (approved):
ACK

PR #153, opened 2023-01-11
Unique Group Ids

sinui0:
This PR builds on #152

It adds the `GroupId` type which uniquely identifies a group within a circuit (old id field rename to index). Additionally, I've added some invariants such as `CircuitId` and `GroupId` must be 1-16 bytes long.

`CircuitId` is no longer a hash of the circuit gates, that was pointless. Circuits are built locally and assumed to be correct based on that.


themighty1 (approved):
ack

PR #156, opened 2023-01-13
Label Sets

sinui0:
It was going to slow me down to split the changes of this PR up, so hopefully this is still legible.

This PR does the following:
1. Introduces `LabelsSet` (aliased as `FullInputLabelsSet` and `ActiveInputLabelsSet`, etc) which is a type to represent a _complete_ set of labels for the io layers of a circuit. It consolidates a lot of validation from around the codebase into this type which should greatly improve safety and probably performance. This type replaces `SanitizedInputLabels` as it serves the same purpose.
2. Decouples the transmission of active input labels from the GC. The `Partial` garbled circuit state no longer includes the generator's `ActiveInputLabels`, those are sent separately. This also paves the way for streaming execution support.
3. Temporarily disables `dualex` in `mpc-aio`, I have another PR coming up for those updates.

### DIFF
--- a/mpc/mpc-aio/src/protocol/garble/exec/mod.rs
+++ b/mpc/mpc-aio/src/protocol/garble/exec/mod.rs
@@ -1,1 +1,1 @@
-pub mod dual;
+//pub mod dual;


themighty1 (approved):
ack

PR #157, opened 2023-01-13
Ghash aio

th4s:
This PR adds the ghash aio layer. It replaces #150.


tls/tls-2pc-aio/benches/prf.rs
```diff
@@ -1,43 +1,43 @@
-use criterion::{black_box, criterion_group, criterion_main, Criterion};
```
themighty1, line 1: Why was this commented out?

themighty1 (approved):
ack

PR #163, opened 2023-01-16
DEAP

sinui0:
This PR implements the simplified variant of DEAP with minimal deviations from standard dual execution.

### DIFF
--- a/mpc/mpc-aio/src/protocol/garble/exec/deap/leader.rs
+++ b/mpc/mpc-aio/src/protocol/garble/exec/deap/leader.rs
@@ -201,7 +201,7 @@ where
         // Also compress follower's circuit to reduce memory footprint before validating it in the next phase
         let gc_cmp = self.backend.compress(gc_evaluated).await?;
 
-        // Commit to output of follower's circuit
+        // Commit to equality check value
         let (commit, leader) = leader.from_compressed_circuit(gc_cmp)?.commit();
 
         self.channel
--- a/mpc/mpc-aio/src/protocol/garble/backend/rayon.rs
+++ b/mpc/mpc-aio/src/protocol/garble/backend/rayon.rs
@@ -93,7 +93,7 @@ impl Compressor for RayonBackend {
     ) -> Result<GarbledCircuit<gc_state::Compressed>, GCError> {
         let (sender, receiver) = oneshot::channel();
         rayon::spawn(move || {
-            _ = sender.send(Ok(circ.to_compressed()));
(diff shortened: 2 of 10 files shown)


themighty1:
I'll have to wait for the write-up but on the first glance this is not what I had in mind.
What I had in mind is that the leader commits to a salted checked value and the follower reveals her checked value. 
Looking at `test_success`, there is sth other than what I had in mind going on, so i'll wait for the write-up first before reviewing the code.

mpc/mpc-core/src/garble/exec/deap/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+        (purported_output, leader, follower)
+    }
+
+    #[test]
+    fn test_success() {
+        let circ = Circuit::load_bytes(ADDER_64).unwrap();
+        let (purported_output, leader, follower) = evaluated_pair(circ.clone());
+
+        let (leader_commit, leader) = leader.commit();
+        let (follower_output, follower) = follower.reveal(leader_commit);
```
themighty1, line 72: This commitment should be salted.

themighty1: I mean the line above `leader.commit();`

mpc/mpc-core/src/garble/exec/deap/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        let (leader_output_values, leader) = leader.decode(follower_output.into()).unwrap();
+
+        let (follower_opening, follower) = follower.open();
```
themighty1, line 76: This opens only the gc? at what stage does the ot get opened?

mpc/mpc-aio/src/protocol/garble/exec/deap/leader.rs
```diff
@@ -0,0 +1,310 @@
...
+        // Evaluate follower's garbled circuit
+        let gc_evaluated = self.backend.evaluate(gc_ev, self.state.ev_labels).await?;
+
+        // Summarize follower's circuit to reduce memory footprint
+        let gc_evaluated_summary = gc_evaluated.get_summary();
+
+        // Also compress follower's circuit to reduce memory footprint before validating it in the next phase
+        let gc_cmp = self.backend.compress(gc_evaluated).await?;
+
+        // Commit to output of follower's circuit
```
themighty1, line 204: It's better to use DualEx terminilogy here: we are committing to the value from the equality test of DualEx.

themighty1 (approved):
looks good, ack

PR #164, opened 2023-01-18
Circuit tests

valpaq:
1) mb it's not clear, why if one gate output is in gate input, no error about this
ofc there will be error in gates, but with another error msg "All output wires must be mapped to gate outputs"
2) if one input gate is in gate output of some cell, error is "All input wires must be mapped to gate inputs:", however it's about overwriting input
3) there is some strange error in test, when I try to add useless gates, added in pr

### DIFF
--- a/mpc/mpc-circuits/src/circuit.rs
+++ b/mpc/mpc-circuits/src/circuit.rs
@@ -784,4 +784,188 @@ mod tests {
         let err = gate.validate().unwrap_err();
         assert!(matches!(err, CircuitError::InvalidCircuit(_)));
     }
+
+    #[test]
+    fn test_strange_error() {
+        let inputs = vec![
+            UncheckedGroup::new(
+                0,
+                "test_input_0".to_string(),
+                "".to_string(),
+                ValueType::Bits,
+                vec![0, 1, 3],
+            ),
+            UncheckedGroup::new(
+                1,
+                "test_input_0".to_string(),
+                "".to_string(),
+                ValueType::Bits,
+                vec![7, 8],
+            ),
+        ];
+        let gates = vec![
+            Gate::Xor {
+                id: 0,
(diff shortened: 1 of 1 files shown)


valpaq:
> Rather than matching on an exact error output, you can just add tests which return _some_ error. The description doesn't matter as much compared to the failed invariant being caught.

I'm not matching, just writing tests and checking error and comparing in my head, and if expected I add matching

valpaq:
last question, in one gate could be different ValueType, e.g.
```
let inputs = vec![UncheckedGroup::new(
    0,
    "test_input_0".to_string(),
    "".to_string(),
    ValueType::Bool,
    vec![0],
),
UncheckedGroup::new(
    1,
    "test_input_1".to_string(),
    "".to_string(),
    ValueType::Bits,
    vec![1],
)
];
let gates = vec![
    Gate::Xor {
        id: 0,
        xref: 0,
        yref: 1,
        zref: 2,
    },
];
```
because shows no error
pls check commit, I think it's enough

valpaq:
> ```shell
> test_overlapping_input_output
> ```
strange, have you pulled last version?
```shell
circuit::tests::test_all_inputs_must_be_connected ... ok
test circuit::tests::test_all_outputs_must_be_connected ... ok
test circuit::tests::test_different_input_value_type ... ok
test circuit::tests::test_doubling_gates_inputs ... ok
test circuit::tests::test_doubling_groups_names_on_input ... ok
test circuit::tests::test_doubling_groups_on_input ... ok
test circuit::tests::test_gate_invariants ... ok
test circuit::tests::test_gates_multiple_outputs ... ok
test circuit::tests::test_overlapping_input_output ... ok
test circuit::tests::test_no_duplicate_group_ids ... ok
test circuit::tests::test_put_input_in_output_gate ... ok
```

PR #166, opened 2023-01-21
Groupless labels

sinui0:
This PR removes the wire id from the label type, and renames labels from `WireLabel` -> `Label`. It also aliases the types in label::state so they can be handled independently without the associated groups.

The latter is a somewhat temporary measure until we reach consensus on type naming and rework the types for hopefully the last time.


themighty1 (approved):
ack

PR #167, opened 2023-01-22
Refactor encoder

sinui0:
This PR builds on #168 and refactors the encoder in `mpc-core`.

Changes:
1. Introduced a trait so we can be generic over encoder if we want
2. Replaced stream_id with u32, this should be more than enough streams and simplifies the reserved stream logic.
3. Removed "domain" functionality, we don't need it.
4. Added `encode_padded` functionality to assist with padding bits on inputs.
5. Introduced flag to reverse the encoding if required (LSB0 -> MSB0)

### DIFF
--- a/mpc/mpc-core/src/garble/label/encoder.rs
+++ b/mpc/mpc-core/src/garble/label/encoder.rs
@@ -63,7 +63,7 @@ pub trait Encoder: Send + Sync {
     fn get_stream(&mut self, stream_id: u32) -> &mut dyn EncoderRng;
 }
 
-/// Encodes wire labels using the ChaCha algorithm.
+/// Encodes wires into labels using the ChaCha algorithm.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


themighty1 (approved):
ack

mpc/mpc-core/src/garble/label/encoder.rs
```diff
@@ -1,19 +1,72 @@
...
-use mpc_circuits::Input;
-use rand::SeedableRng;
+use mpc_circuits::{Input, WireGroup};
+use rand::{CryptoRng, RngCore, SeedableRng};
 use rand_chacha::ChaCha20Rng;
 
-use super::{Delta, FullEncodedInput};
+use crate::Block;
 
-/// Encodes wire labels using the ChaCha algorithm and a global offset (delta).
```
themighty1, line 9: better "Encodes wires into labels" ... (to avoid ambiguity)

PR #171, opened 2023-01-24
Refactor 2PC GHASH

sinui0:
This PR moves the ghash impl into the `universal-hash` crate. The core logic should be untouched.

**Changes**
1. Updated interface to support initialization prior to knowing the key share
2. Introduced the `UniversalHash` trait which is a common interface between GHASH and Poly1305
3. Renamed a couple variables such as `max_message_length` -> `max_block_count` as the message vs block naming was a bit confusing.
4. Automatically pad the message if it isn't a multiple of the block size
5. Automatically compute new shares if the block count increases.


themighty1 (approved):
ack

PR #172, opened 2023-01-25
Garbled Circuit ZKP

sinui0:
This PR implements zero-knowledge proofs using garbled circuits.

We need this for our decryption protocol (for now).

### DIFF
--- a/mpc/mpc-core/src/garble/circuit.rs
+++ b/mpc/mpc-core/src/garble/circuit.rs
@@ -177,7 +177,7 @@ pub struct GarbledCircuit<S: State> {
 /// "opens" the GC.
 /// We rely on the property of the "half-gates" garbling scheme that given the input
 /// label pairs and the delta, a GC will always be generated deterministically.
-/// We assume that the evaluator is already in posession of all active input labels.
+/// We assume that the evaluator is already in possession of all active input labels.
 ///
 /// Note that instead of `input_decoding`, the circuit generator could just provide
 /// his actual input to the circuit. But this would have a slightly larger cost when


mpc/mpc-aio/src/protocol/garble/exec/zk/prover.rs
```diff
@@ -0,0 +1,264 @@
...
+
+        // Reveal output to follower
+        let (commit_opening, output) = prover.reveal();
+
+        self.channel
+            .feed(GarbleMessage::CommitmentOpening(commit_opening.into()))
+            .await?;
+
+        self.channel
+            .send(GarbleMessage::Output(output.into()))
```
themighty1, line 259: what's the purpose of this? these are the same labels as in CommitmentOpening above

mpc/mpc-aio/src/protocol/garble/exec/zk/verifier.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub async fn verify(mut self) -> Result<Vec<OutputValue>, GCError> {
+        // Open our circuit to the Prover
+        let (opening, verifier) = self.state.verifier.open();
+
+        self.channel
+            .send(GarbleMessage::CircuitOpening(opening.into()))
+            .await?;
+
+        // Receive opening to output commitment
+        let commitment_msg = expect_msg_or_err!(
```
themighty1, line 199: should be `opening_msg`

themighty1 (commented):
also a q from discord about input_decoding

mpc/mpc-aio/src/protocol/garble/exec/zk/prover.rs
themighty1: Do you plan it as a future optimization that the prover sends the actual output instead of output labels and then verifier reconstructs the commitment?

themighty1 (approved):
ack

PR #175, opened 2023-01-26
DualEx option to skip equality check

sinui0:
This PR adds the option to skip (defer) the equality check protocol in dual execution. This will be used in the PRF to output the symmetric key labels for the AEAD.

I hardcoded this to have both parties send output labels commitments for good measure. I don't think it is necessary but it can't hurt.

### DIFF
--- a/mpc/mpc-aio/src/protocol/garble/exec/dual.rs
+++ b/mpc/mpc-aio/src/protocol/garble/exec/dual.rs
@@ -213,7 +213,7 @@ where
     /// Execute dual execution protocol without the equality check
     ///
     /// This can be used when chaining multiple circuits together. Neither party
-    /// reveals the output label decoding.
+    /// reveals the output label decoding information.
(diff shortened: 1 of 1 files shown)


mpc/mpc-aio/src/protocol/garble/exec/dual.rs
```diff
@@ -209,6 +209,56 @@ where
 
         Ok(gc_evaluated.into_summary())
     }
+
+    /// Execute dual execution protocol without the equality check
+    ///
+    /// This can be used when chaining multiple circuits together. Neither party
+    /// reveals the output label decoding.
```
themighty1, line 216: I think we should only use `decoding` in names but when explaining in comments we should always say "decoding information".  We use the `decoding` shorthand internally but outside of our circle this will not be understood.

themighty1 (approved):
ack

PR #177, opened 2023-01-27
DualEx use OT Factory

sinui0:
This PR refactors DE and DEAP to use OT factories instead of requiring the OT dependencies to be injected at construction time.

It also introduces configuration structs for both.


themighty1 (approved):
ack

PR #178, opened 2023-01-27
DualEx Traits

sinui0:
This PR builds on #177 and introduces convenience traits for dual execution. I also split up some modules for readability.

PR #179, opened 2023-01-27
DualEx Factory

sinui0:
This PR builds on #178 , it implements generic factories for DE and DEAP. It also moves the circuit reference into the config structs.


themighty1 (approved):
ack on the code but I don't understand the rational behind having a factory for this.

Could you elaborate how having a factory in this context is more beneficial vs creating classes directly?

PR #182, opened 2023-01-29
Free-XOR ops

sinui0:
This PR implements the BitXor ops for full and active labels. Thanks to free-XOR, we can locally compute the XOR of encoded data. This will come in handy quite a bit for us for removing one-time pads from output labels without having to define a new circuit or implement "intermediate groups" yet.


themighty1 (approved):
ack

PR #184, opened 2023-01-31
Generic share conversion

th4s:
- share conversion is generic over field trait
- implement OT for arbitrary message length
- add the field P256
- adapt all usages and tests

### DIFF
--- a/mpc/share-conversion-core/src/fields/mod.rs
+++ b/mpc/share-conversion-core/src/fields/mod.rs
@@ -121,10 +121,18 @@ mod tests {
 
     pub fn test_field_bit_ops<T: Field>() {
         let mut a = vec![false; T::BIT_SIZE as usize];
-        a[T::BIT_SIZE as usize - 1] = true;
+        let mut b = vec![false; T::BIT_SIZE as usize];
 
-        let out = T::from_bits_msb0(&a);
-        assert_eq!(out, T::one());
-        assert_eq!(out.get_bit_msb0(T::BIT_SIZE - 1), true);
+        a[T::BIT_SIZE as usize - 1] = true;
+        b[0] = true;
+
+        let a = T::from_bits_msb0(&a);
+        let b = T::from_bits_msb0(&b);
+
+        assert_eq!(a, T::one());
+        assert_eq!(a.get_bit_msb0(T::BIT_SIZE - 1), true);
+
+        assert_eq!(b, T::one() * T::two_pow(T::BIT_SIZE - 1));
+        assert_eq!(b.get_bit_msb0(0), true);
     }
 }


mpc/share-conversion-core/src/lib.rs
```diff
@@ -1,23 +1,200 @@
...
+    Self: Sized,
+{
+    type Inner: Field;
+    type Output: ShareConvert<Inner = Self::Inner, Output = Self>;
+
+    /// Create a new instance
+    fn new(share: Self::Inner) -> Self;
+
+    /// Converts '&self' into choices needed for the receiver input to an oblivious transfer.
+    /// The choices are in the "least-bit-first" order.
```
themighty1, line 41: Either the comment is wrong or the code is wrong. The way I read the code is:
- take the MSB bit and push it to `out`
- repeat for the rest of the bits until LSB bit is pushed
the result is that the first bit of `out` is MSB. This is not "least-bit-first" as the comment says.

mpc/share-conversion-core/src/m2a.rs
```diff
@@ -0,0 +1,58 @@
...
+    ) -> Result<(AddShare<T>, OTEnvelope<T>), ShareConversionCoreError> {
+        // create random masks
+        let mut masks: Vec<T> = vec![T::zero(); T::BIT_SIZE as usize];
+        masks.iter_mut().for_each(|x| *x = T::rand(rng));
+
+        let t0: Vec<T> = masks.clone();
+
+        // we multiply `self.inner()` with 2^i and add a mask
+        let mut t1 = vec![T::zero(); T::BIT_SIZE as usize];
+        for (k, t) in t1.iter_mut().enumerate() {
```
themighty1, line 29: the comment says it should be ```T::two_pow(k)```

mpc/share-conversion-core/src/fields/gf2_128.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+impl Mul for Gf2_128 {
+    type Output = Self;
+
+    /// Galois field multiplication of two 128-bit blocks reduced by the GCM polynomial
+    fn mul(self, rhs: Self) -> Self::Output {
+        /// R is the GCM polynomial in big-endian. In hex: "000000000000000000000000000000E1"
+        const R: u128 = 135;
```
themighty1, line 54: A note to self:
After other Qs are resolved I want to examine if this is a correct algorithm endianess-wise.

mpc/mpc-aio/src/protocol/ot/kos/receiver.rs
```diff
@@ -110,6 +115,50 @@ impl ObliviousReceive<bool, Block> for Kos15IOReceiver<r_state::RandSetup> {
     }
 }
 
+// The idea is to send AES encryption keys in the OT, which can then later be used by the receiver
```
themighty1, line 118: // The idea for obliviously receiving a vec of Blocks is for the sender to send an AES encryption key
// using OT, which can then later be used by the receiver to decrypt an arbitrary long message, which
// is sent shortly after the OT. This way we extend our OT from 128-bit maximum message length to an
// unlimited message length.

mpc/mpc-aio/src/protocol/ot/kos/receiver.rs
```diff
@@ -110,6 +115,50 @@ impl ObliviousReceive<bool, Block> for Kos15IOReceiver<r_state::RandSetup> {
...
+// to decrypt arbitrary long messages, which are sent shortly after the OT. This way we extend our
+// OT from 128-bit maximum message length to an unlimited message length
+#[async_trait]
+impl ObliviousReceive<bool, Vec<Block>> for Kos15IOReceiver<r_state::RandSetup> {
+    async fn receive(&mut self, choices: Vec<bool>) -> Result<Vec<Vec<Block>>, OTError> {
+        // Receive AES encryption keys from OT
+        let keys = ObliviousReceive::<bool, Block>::receive(self, choices.clone()).await?;
+
+        // Receive ciphertexts
```
themighty1, line 127: // Receive a pair of ciphertexts for each choice bit

mpc/mpc-aio/src/protocol/ot/kos/receiver.rs
```diff
@@ -110,6 +115,50 @@ impl ObliviousReceive<bool, Block> for Kos15IOReceiver<r_state::RandSetup> {
...
+        let ExtSenderEncryptedPayload { mut ciphertexts } = expect_msg_or_err!(
+            self.channel.next().await,
+            OTMessage::ExtSenderEncryptedPayload,
+            OTError::Unexpected
+        )?;
+
+        // Decrypt ciphertexts with keys
```
themighty1, line 134: // Decrypt one of the ciphertexts from each pair with a corresponding key

mpc/mpc-aio/src/protocol/ot/kos/sender.rs
```diff
@@ -120,6 +127,87 @@ impl ObliviousSend<[Block; 2]> for Kos15IOSender<s_state::RandSetup> {
     }
 }
 
+// The idea is to send AES encryption keys in the OT, which can then later be used by the receiver
```
themighty1, line 130: // The idea for obliviously sending a vec of Blocks is to send an AES encryption key using OT, which
// can then later be used by the receiver to decrypt an arbitrary long message, which is sent shortly
// after the OT. This way we extend our OT from a 128-bit maximum message length to an unlimited message
// length.

mpc/mpc-aio/src/protocol/ot/kos/sender.rs
```diff
@@ -120,6 +127,87 @@ impl ObliviousSend<[Block; 2]> for Kos15IOSender<s_state::RandSetup> {
...
+            let (cipher1, cipher2) = (
+                Aes128::new(&key1.inner().to_be_bytes().into()),
+                Aes128::new(&key2.inner().to_be_bytes().into()),
+            );
+            cipher1.encrypt_blocks(&mut inputs[k][0]);
+            cipher2.encrypt_blocks(&mut inputs[k][1]);
+            keys.push([key1, key2]);
+        }
+
+        // Send keys in OT
```
themighty1, line 168: // Send keys using OT

mpc/mpc-aio/src/protocol/ot/kos/sender.rs
```diff
@@ -120,6 +127,87 @@ impl ObliviousSend<[Block; 2]> for Kos15IOSender<s_state::RandSetup> {
...
+        ObliviousSend::<[Block; 2]>::send(self, keys).await?;
+
+        // Convert input back to blocks
```
themighty1, line 171: // Convert inputs back to block

mpc/mpc-core/src/msgs/ot.rs
```diff
@@ -112,6 +113,16 @@ pub struct ExtSenderReveal {
     pub offset: usize,
 }
 
+/// We use this message, when we want to send data which is longer than 128 bits
```
themighty1, line 116: /// We use this message when we want to send data which is longer than 128 bits

mpc/share-conversion-aio/src/conversion/msgs.rs
```diff
@@ -0,0 +1,35 @@
+use crate::ShareConversionError;
+use share_conversion_core::fields::Field;
+use utils_aio::Channel;
+
+/// A channel used for messaging of conversion protocols
```
themighty1, line 5: /// A channel used by conversion protocols for messaging

mpc/share-conversion-core/src/a2m.rs
```diff
@@ -0,0 +1,91 @@
...
+            .fold(T::zero(), |acc, i| acc + *i);
+
+        // the inverse of the random share will be the multiplicative share for the sender
+        let mul_share = MulShare::new(random.inverse());
+
+        // decompose the share into a sum of components, e.g. if the share is 10110, we decompose it into
+        // 0 + 10 + 100 + 0000 + 10000
+        let components: Vec<T> = (0..T::BIT_SIZE)
+            .map(|k| {
+                // we extract a bit of `self.inner()` in position `i` (counting from the left) and
```
themighty1, line 45: // we extract a bit of `self.inner()` in position `k` (counting from the left) 
// then left-shift that bit by `k`;

mpc/share-conversion-core/src/fields/gf2_128.rs
```diff
@@ -0,0 +1,195 @@
...
+        Self(self.0 ^ rhs.0)
+    }
+}
+
+impl Mul for Gf2_128 {
+    type Output = Self;
+
+    /// Galois field multiplication of two 128-bit blocks reduced by the GCM polynomial
+    fn mul(self, rhs: Self) -> Self::Output {
+        /// R is the GCM polynomial in big-endian. In hex: "000000000000000000000000000000E1"
```
themighty1, line 53: /// R is the GCM "special element" (see section 2.5 of "The Galois/Counter Mode of Operation (GCM)")
/// in big-endian. In hex: "000000000000000000000000000000E1"

mpc/share-conversion-core/src/fields/gf2_128.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    #[test]
+    fn test_gf2_128_bit_ops() {
+        test_field_bit_ops::<Gf2_128>();
+    }
+
+    #[test]
+    fn test_gf2_128_mul() {
+        // Naive mutltiplication is the same here
```
themighty1, line 163: // Naive multiplication is the same here

themighty1 (changes requested):
nice work.
I copy pasted from my git diff some comment improvements and typos
Still need to look into endianess issue

mpc/mpc-aio/src/protocol/ot/kos/sender.rs
themighty1: yes, sry my typo

mpc/share-conversion-core/src/fields/gf2_128.rs
themighty1: This algo treats all field elements as if their bit order is MSB ?
I'm ok to keep it if there is a concrete benefit of doing that. 

The downsides are:
-  it departs from the GCM paper's algo and makes it harder to audit
- this seems like a unnecessary step. Normally we would take a block of the plaintext and feed it to ghash as-is. With this approach we need to first reverse the bits of the plaintext and then feed it to ghash.

themighty1: > a block of the plaintext 
sry, I meant to say a block of ciphertext. By "block" I mean `vec![u8; 16]` - the bytes which are outputted by the AES cipher.

mpc/share-conversion-core/src/a2m.rs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        let mut b1 = vec![T::zero(); T::BIT_SIZE as usize];
+        for (k, ((b, c), m)) in b1
+            .iter_mut()
+            .zip(components.iter())
+            .zip(masks.iter())
+            .enumerate()
+        {
+            *b = ((*c + T::two_pow(T::BIT_SIZE - k as u32 - 1)) * random) + *m;
```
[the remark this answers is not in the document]
themighty1: It makes it harder to audit the correstness of this algo, since the paper  says
ddd

With this approach we have the extra burden of explaining why it is ok to make this modification

themighty1 (approved):
ACK

PR #186, opened 2023-02-03
Key Exchange

th4s:
This PR adds the code for the key exchange using share-conversion and garbled circuit dual execution


key-exchange/src/exchange.rs
```diff
@@ -0,0 +1,523 @@
...
+
+        // Check that the output of the XOR circuit is equal to 0, i.e. PMS1 == PMS2
+        if *xor_output != vec![0_u8; 32] {
+            return Err(KeyExchangeError::CheckFailed);
+        }
+
+        // Turn output into labels and return them
+        // We only need half the labels because we only need one of the two PMS values (which we
+        // know are equal to each other)
+        // Since, there are only 2 output labels, we can just take the first one
```
themighty1, line 269: you probably meant "only 2 outputs" or "only 2 output label sets"

themighty1 (approved):
ack, gw

PR #190, opened 2023-02-09
Point addition

th4s:
This PR implements point addition for P256.

### DIFF
--- a/point-addition/src/conversion.rs
+++ b/point-addition/src/conversion.rs
@@ -22,10 +22,28 @@ where
     a2m_converter: T,
     /// Performs conversion of multiplication to addition
     m2a_converter: U,
-    /// A flag which is used to indicate the party, which has to do an inversion operation
-    negate: bool,
+    /// Indicates which role this converter instance will fulfill
+    role: Role,
     /// PhantomData used for the underlying elliptic curve field
     _field: std::marker::PhantomData<V>,
+}
+
+/// The role: either Leader or Follower
+///
+/// Follower needs to perform an inversion operation on the point during point addition
+#[derive(Debug, Clone, Copy)]
+pub enum Role {
+    Leader,
+    Follower,
+}
+
+impl Role {
+    fn adapt_point<V: Field>(&self, [x, y]: [V; 2]) -> [V; 2] {
+        match self {
+            Role::Leader => [x, y],
+            Role::Follower => [-x, -y],
+        }
+    }
 }
 
 impl<T, U, V> Converter<T, U, V>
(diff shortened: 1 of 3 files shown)


point-addition/src/conversion.rs
```diff
@@ -0,0 +1,91 @@
+//! This module implements the conversion from the sum of two elliptic curve points to the sum of
```
themighty1, line 1: I dont think the word "conversion" is fitting for this module.

E.g. M2A means "multiplicative shares to additive shares" conversion. Multiplicative shares is what we have in the beginning of the protocol. 

If we say "conversion from the sum" that would mean that "sum" is sth that was originally available, which is not accurate.

Instead I suggest to use the wording similar to what is used in lib.rs

//! This module implements a secure two-party computation protocol for adding two private EC points and secret-sharing the resulting x coordinate (the shares are field elements of the field underlying the elliptic curve).
//! This protocol has semi-honest security.

point-addition/src/conversion.rs
```diff
@@ -0,0 +1,91 @@
...
+//! two field elements of the field underlying the elliptic curve
+
+use super::{PointAddition, PointAdditionError};
+use async_trait::async_trait;
+use p256::EncodedPoint;
+use share_conversion_aio::{AdditiveToMultiplicative, MultiplicativeToAdditive};
+use share_conversion_core::fields::{p256::P256, Field};
+
+/// The instance used for converting the curve points
```
themighty1, line 10: "adding" the curve points

point-addition/src/conversion.rs
```diff
@@ -0,0 +1,91 @@
...
+pub struct Converter<T, U, V>
+where
+    T: AdditiveToMultiplicative<V>,
+    U: MultiplicativeToAdditive<V>,
+    V: Field,
+{
+    /// Performs conversion of addition to multiplication
```
themighty1, line 17: Here and below "conversion" is correct, since we do A2M conversion for the field elements

point-addition/src/conversion.rs
```diff
@@ -0,0 +1,91 @@
...
+            a2m_converter,
+            m2a_converter,
+            negate,
+            _field: std::marker::PhantomData,
+        }
+    }
+
+    /// Perform the conversion of P = A + B => P_x = a + b
+    ///
+    /// This will convert an elliptic curve point addition to an additive sharing in the underlying
```
themighty1, line 45: While I understand what you mean by "convert addition to ...", for outsiders it will be hard to understand because "addition" is a process of adding. 
It may be understood as if "instead of performing addition , we perform additive sharing"

point-addition/src/conversion.rs
```diff
@@ -0,0 +1,91 @@
+//! This module implements the conversion from the sum of two elliptic curve points to the sum of
+//! two field elements of the field underlying the elliptic curve
```
themighty1, line 2: I suggest to also prime the reader (or potential contributors who try to understand our code now) by adding

//! The protocol is described in https://docs.tlsnotary.org/protocol/notarization/key_exchange.html

(at least temporarily, we will later have a dedicated point_addition page and will update this comment)

themighty1 (approved):
ACK
Let's some to an agreemtn on the wording. I understand that it is tempting to use "conversion"  but I'm not convinced that it is accurate to use that term.

point-addition/src/conversion.rs
themighty1: thank you

PR #191, opened 2023-02-10
Unify Dualex API

sinui0:
This PR refactors our API for dual execution, unifying standard DE and DEAP's API so they can be used interchangeably.

Additionally, I've added types to assist with deferring DEAP's equality check to another context via a sink. This will help make our other protocols generic over any type of dual execution variant we use in the future.


themighty1 (approved):
ACK, had to pause in one place, so pushed a comment for clarity, feel free to modify it.

PR #193, opened 2023-02-10
ZK refactor + Defer

sinui0:
This PR refactors our GC ZK impl and adds types for deferral like in #191 

Changes:
1. Changed `Verify` trait to accept `expected_output`
2. Added summary methods to both traits
3. Updated both `Prover` and `Verifier` constructors to accept configs and OT factories
4. Fixed oversight where `Verifier` wasn't opening OTs to the `Prover`, tests passed because they are using MockOT but this would have failed in integration tests.
5. Added `Deferred` types.


themighty1 (approved):
ACK with comments and q

PR #194, opened 2023-02-10
ZK Factory

sinui0:
This PR builds on #193 and implements factories for constructing instances of `Prover` and `Verifier`


themighty1 (approved):
ack

PR #198, opened 2023-02-21
BitOrder Support

sinui0:
This PR updates the mpc-circuits model to support different bit orders (LSB0 vs MSB0).

Unfortunately all the low level circuits I wrote are LSB0, but I'm not going to tackle that in this PR.

Changes:
1. Fix bug in circuit builder API which doesn't preserve output wire order
2. Adds `BitOrder` to the circuit model
3. Updates our circuit binaries (AES, SHA256, ADDER64) to be MSB0 and BE.
4. Add "reverse" flag to our cli helper for parsing bristol circuits
5. Adds `BitOrder` to the `ChaChaEncoder` so we can configure the bitorder of the stream transcript if needed.
6. Temporarily disables crates in the `tls` workspace. I have many major breaking changes coming for them, so updating them to support the new bit order is a waste of time.

### DIFF
--- a/mpc/mpc-circuits/src/bin/parse.rs
+++ b/mpc/mpc-circuits/src/bin/parse.rs
@@ -15,9 +15,11 @@ struct Args {
     name: String,
     #[clap(short)]
     version: String,
-    /// This flag reverses the circuit the circuit input and output wires, ie flips the bit order.
+    /// This flag reverses the circuit input and output wires, ie flips the bit order
+    /// of the circuit.
     ///
-    /// Any byte values will also be reversed, ie flipping the endianness.
+    /// Any byte values will also be reversed, ie flipping the endianness. Little-endian
+    /// inputs/outputs will become big-endian and vice versa.
     #[clap(short, default_value = "false")]
     reverse: bool,
     /// Path to bristol fashion circuit
@@ -54,6 +56,7 @@ fn main() {
     println!("Successfully processed {}", &args.name);
 }
 
+// Reverses the input and output wires of a circuit.
 fn reverse_bristol(circ: &Circuit) -> Arc<Circuit> {
     let id = circ.id().clone().to_string();
     let mut builder = CircuitBuilder::new(
@@ -83,6 +86,7 @@ fn reverse_bristol(circ: &Circuit) -> Arc<Circuit> {
 
     let original_circ = builder.add_circ(&circ);
 
+    // For each input, create a new input with reversed wires.
     for (idx, input) in new_inputs.iter().enumerate() {
         let input_wires = input[..].iter().cloned().collect::<Vec<_>>();
         let original_input_wires = original_circ.input(idx).unwrap()[..]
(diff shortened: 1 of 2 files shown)


mpc/mpc-circuits/src/bin/parse.rs
```diff
@@ -1,41 +1,126 @@
...
+    let original_circ = builder.add_circ(&circ);
+
+    for (idx, input) in new_inputs.iter().enumerate() {
+        let input_wires = input[..].iter().cloned().collect::<Vec<_>>();
+        let original_input_wires = original_circ.input(idx).unwrap()[..]
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>();
+
+        for (feed, sink) in input_wires.iter().zip(original_input_wires.iter().rev()) {
```
themighty1, line 93: I have a hard time picturing what this loop does. Could you add a comment?

mpc/mpc-circuits/src/bin/parse.rs
```diff
@@ -1,41 +1,126 @@
...
+        .collect::<Vec<_>>();
+
+    for (idx, output) in new_outputs.iter().enumerate() {
+        let output_wires = output[..].iter().cloned().collect::<Vec<_>>();
+        let original_output_wires = original_circ.output(idx).unwrap()[..]
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>();
+
+        for (feed, sink) in original_output_wires.iter().zip(output_wires.iter().rev()) {
```
themighty1, line 120: and also a comment here is needed.

mpc/mpc-circuits/src/builder.rs
```diff
@@ -497,6 +504,32 @@ impl CircuitBuilder<Outputs> {
...
     pub fn build_circuit(mut self) -> Result<Arc<Circuit>, BuilderError> {
+        let output_wire_ids = self
+            .0
+            .outputs
+            .iter()
+            .map(|output| output.output.wires.clone())
+            .flatten()
+            .collect::<HashSet<usize>>();
+
+        let mut output_wire_connections: HashMap<usize, usize> = HashMap::default();
```
themighty1, line 515: Could you add a comment about what this block does?

mpc/mpc-circuits/src/builder.rs
```diff
@@ -280,6 +283,7 @@ pub struct Outputs {
     id: String,
     description: String,
     version: String,
+    bit_order: BitOrder,
     inputs: Vec<InputHandle>,
     gates: Vec<GateHandle>,
     conns: HashMap<usize, usize>,
```
themighty1, line 289: Could you add a comment here about what `conns` is?

themighty1 (changes requested):
Im gonna need some more comments to better understand the fundamentals and then I'll give it a review.

themighty1 (approved):
ack

PR #201, opened 2023-02-22
2PC Block Cipher

sinui0:
This PR builds on #199 and provides a 2PC block cipher generic over the cipher.

Features

1. Encrypt a message provided by one of the parties
2. Compute additive shares of a ciphertext (we'll use this to compute shares of the hash key in AEAD)

### DIFF
--- a/cipher/block-cipher/src/cipher.rs
+++ b/cipher/block-cipher/src/cipher.rs
@@ -93,7 +93,7 @@ where
         let encoder = self
             .state
             .encoder
-            .clone()
+            .as_mut()
             .ok_or(BlockCipherError::EncoderNotSet)?;
 
         // Compute instance id
@@ -323,7 +323,7 @@ where
             Role::Leader => {
                 let mask_leader_input = mask_leader_input
                     .to_value(mask.clone())
-                    .expect("mask is valid length");
+                    .expect("mask should be valid length");
 
                 (
                     vec![text_input.clone(), mask_leader_input.clone()],
--- a/cipher/block-cipher/src/lib.rs
+++ b/cipher/block-cipher/src/lib.rs
@@ -43,8 +43,8 @@ pub enum BlockCipherError {
 
 #[derive(Clone)]
 pub struct BlockCipherLabels {
-    key_full: FullLabels,
-    key_active: ActiveLabels,
+    pub key_full: FullLabels,
+    pub key_active: ActiveLabels,
 }
 
 #[async_trait]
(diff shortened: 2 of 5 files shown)


cipher/block-cipher/src/cipher.rs
```diff
@@ -0,0 +1,377 @@
...
+impl<C, DEF, DE> BlockCipher<C> for DEBlockCipher<C, DEF, DE>
+where
+    C: BlockCipherCircuitSuite,
+    DEF: AsyncFactory<DE, Config = DualExConfig, Error = GCFactoryError> + Send,
+    DE: DEExecute + Send,
+{
+    /// Sets the key input labels for the block cipher.
+    ///
+    /// * `labels`: The labels to use for the key input.
+    fn set_keys(&mut self, labels: BlockCipherLabels) {
```
themighty1, line 71: this should be pub

cipher/block-cipher/src/cipher.rs
```diff
@@ -0,0 +1,377 @@
...
+        self.state.labels = Some(labels);
+    }
+
+    /// Sets the encoder used to generate the input labels
+    /// used during 2PC.
+    fn set_encoder(&mut self, encoder: Arc<Mutex<ChaChaEncoder>>) {
```
themighty1, line 77: this should be pub

themighty1 (changes requested):
I would like to see a test which shows how this will be used in TLSNotary to make it easier to follow.
Then I'll review again

cipher/block-cipher/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+    };
+    use mpc_circuits::{BitOrder, Value};
+    use mpc_core::garble::Encoder;
+
+    use super::*;
+
+    pub type MockDEBlockCipherLeader<C> = DEBlockCipher<C, MockDualExFactory, MockDualExLeader>;
+    pub type MockDEBlockCipherFollower<C> = DEBlockCipher<C, MockDualExFactory, MockDualExFollower>;
+
+    pub fn create_labels<C: BlockCipherCircuitSuite>(
```
themighty1, line 107: I would expect to see here how the output labels from one GC execution become the key labels for block cipher like it will be in TLSNotary. Is there anything special that the encoder have to be aware of? This will help me see the big picture.

themighty1 (changes requested):

themighty1 (commented):
ack for now, will give a thorough review later.

themighty1 (approved):

PR #204, opened 2023-03-01
2PC AES GCM

sinui0:
This PR builds on #201, it implements 2PC AES GCM using our block cipher, stream cipher and ghash impls.


aead/src/aes_gcm/follower.rs
```diff
@@ -0,0 +1,245 @@
...
+        let tag_share = self
+            .compute_tag_share(explicit_nonce, aad, ciphertext.clone())
+            .await?;
+
+        // Send tag share to leader
+        self.channel
+            .send(AeadMessage::TagShare(tag_share.into()))
+            .await?;
+
+        // Expect tag share from leader
```
themighty1, line 113: In what situation will the Notary want to see the tag share?

themighty1 (approved):
ACK, gw, will do a deeper review later on.

PR #207, opened 2023-03-09
Moar Bitvec Utils

sinui0:
Sorry, I somehow ended up bikeshedding this.

This PR improves our bitvec utils to be more generic, and include implementations for native uint types. I also introduced some macro impl blocks to reduce duplication. Now any iterator of any size uint can be converted into a bit iterator and vice versa.

With these utilities implemented, it gives me some ideas for improving the mpc-circuits API quite a bit. For another day..


th4s (approved):
ack

themighty1 (approved):
ack

PR #208, opened 2023-03-13
added notarized-session type

themighty1:
This PR contains the types needed to "manually" construct the `NotarizedSession` type
The types were extracted from https://github.com/tlsnotary/tlsn/pull/173 and renamed to match the excalidraw diagram.
No functional changes were made.


notarized-session/src/session_data.rs
```diff
@@ -0,0 +1,33 @@
...
+        &self.handshake_data
+    }
+
+    pub fn transcript(&self) -> &Transcript {
+        &self.transcript
+    }
+
+    pub fn commitments(&self) -> &Vec<Commitment> {
+        &self.commitments
+    }
```
th4s, line 32: Replace with:
```rust
    pub fn commitments(&self) -> &[Commitment] {
        self.commitments.as_slice()
    }

```
The advantages are:
  - no double indirection. `&Vec` is a pointer to a pointer to data on the heap.
  - `&[Commitment]` is generic over the memory structure. Everything which implements slicing.

notarized-session/src/handshake_data.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Misc TLS handshake data which the User committed to before the User and the Notary engaged in 2PC
+/// to compute the TLS session keys
+///
+/// The User should not reveal `tls_cert_chain` because the Notary would learn the webserver name
+/// from it. The User also should not reveal `sig_ke_params` to the Notary, because
+/// for ECDSA sigs it is possible to derive the pubkey from the sig and then use that pubkey to find out
+/// the identity of the webserver.
+//
+/// Note that there is no need to commit to the ephemeral key because it will be signed explicitely
```
th4s, line 15: typo: explicitly

th4s (approved):
Nice :rocket:

One minor thing: there are several places, where it is better to return a `&[T]` instead of a `&Vec<T>`.

PR #209, opened 2023-03-16
2PC HMAC-SHA256 PRF

sinui0:
This PR builds on #204 and implements the 2PC HMAC-SHA256 PRF used in TLS1.2.

### DIFF
--- a/prf/hmac-sha256/src/leader.rs
+++ b/prf/hmac-sha256/src/leader.rs
@@ -14,11 +14,11 @@ use mpc_core::garble::{
 };
 use utils_aio::factory::AsyncFactory;
 
-use crate::{circuits, PRFLeader, State};
+use crate::{circuits, PRFLead, State};
 
 use super::PRFError;
 
-pub struct DEPRFLeader<DEF, DE>
+pub struct PRFLeader<DEF, DE>
 where
     DEF: AsyncFactory<DE, Config = DualExConfig, Error = GCFactoryError>,
     DE: DEExecute + Send,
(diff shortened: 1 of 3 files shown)


themighty1 (approved):
ack, gw!

prf/hmac-sha256-core/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+});
+pub static SESSION_KEYS: Lazy<Arc<Circuit>> = Lazy::new(|| {
+    Circuit::load_bytes(std::include_bytes!("../circuits/bin/session_keys.bin")).unwrap()
+});
+pub static CF_VD: Lazy<Arc<Circuit>> = Lazy::new(|| {
+    Circuit::load_bytes(std::include_bytes!("../circuits/bin/cf_verify_data.bin")).unwrap()
+});
+pub static SF_VD: Lazy<Arc<Circuit>> = Lazy::new(|| {
+    Circuit::load_bytes(std::include_bytes!("../circuits/bin/sf_verify_data.bin")).unwrap()
+});
```
th4s, line 25: Can we have some comments here, what these circuits are?

prf/hmac-sha256/src/leader.rs
```diff
@@ -0,0 +1,208 @@
...
+    exec::dual::{DualExConfig, DualExConfigBuilder},
+    ChaChaEncoder,
+};
+use utils_aio::factory::AsyncFactory;
+
+use crate::{circuits, PRFLeader, State};
+
+use super::PRFError;
+
+pub struct DEPRFLeader<DEF, DE>
```
th4s, line 21: Can we change this name to `PRFLeader`? Same for follower.

th4s (approved):
Good work :rocket:, only some small things. Can we remove `tls/{tls-2pc-aio, tls-2pc-core, tls-circuits}`?

prf/hmac-sha256/src/leader.rs
th4s: We could name the trait `PRFLead`?  Only asked because it will be a pain to type this :smile:

Something different: I encountered the pattern of naming the trait exactly as the struct/implementer several times. I think this is a bad idea, because it leads to more confusion especially for people not familiar with the code.

PR #210, opened 2023-03-22
Reorg MPC crates

sinui0:
This PR does some shuffling and cleaning of our MPC code. I've split up `mpc-core` and `mpc-aio`, separating the different protocols into their own crates.

There are no functional changes here, with the exception of `key-exchange` which apparently got broken at some point but wasn't detected because it's not in our CI. I introduced fixes to get its tests passing again, however I also noticed that its circuits are different bit order than the rest of our code, which we'll need to change at some point (and add the crate to CI).


themighty1 (approved):
ACK

PR #212, opened 2023-03-30
mpc-circuits-v2

sinui0:
This PR completely reimplements our `mpc-circuits` crate. Dependent crates are currently broken, this PR will be the first of a handful in order to incorporate these changes.

I put in some effort to document the API, so don't forget to check out the rustdocs if you're trying to understand something. I wouldn't burn a whole lot of time reviewing the macro code (unless you're curious).

I believe this is a considerable simplification in the API, implementation and abstractions. My big mistake with v1 was the `Input/Output` types leaking all over the place and not leveraging the type system nearly enough. Now, a lot of things are statically checked and I hope you'll agree it is now more ergonomic.

I expect these changes to cut out thousands of lines of code in dependent crates, mostly due to the builder API and having static types available.

It will require more tests, but for now I believe it is sufficient to proceed with it as is.

### DIFF
--- a/mpc/mpc-circuits-macros/src/trace.rs
+++ b/mpc/mpc-circuits-macros/src/trace.rs
@@ -132,7 +132,7 @@ pub(crate) fn trace_impl(args: TokenStream, item: TokenStream) -> TokenStream {
                     _ => panic!("Unsupported argument type: {:?}", arg.to_token_stream()),
                 };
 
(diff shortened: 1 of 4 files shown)


themighty1 (approved):
thanks, this approach looks clean

th4s (approved):
good work :rocket:

PR #216, opened 2023-04-17
Adapt share-conversion to new OT interface

th4s:
This PR adapts `mpc/share-conversion` and `actors/actor-share-conversion`to remove the OT factory.


themighty1 (approved):
ack

PR #217, opened 2023-04-18
remove Result from spawn method

sinui0:
This PR refactors our async backend abstraction slightly by removing the `Result` it returns for handling canceled errors. This sort of error doesn't need to be exposed in the API and can just cause a panic instead as it would indicate an unrecoverable bug. Eg. the thread panicking. The function passed into the thread can of course still be fallible, but this way it doesn't force it to implement `From<oneshot::Canceled>`


th4s (approved):
ack

themighty1:
ack

PR #218, opened 2023-04-18
Refactor mpc-core

sinui0:
This PR is one of a series of refactors for our garbled circuit crates.

# Changes
* Embraces serde for all things serialization
  * Removes intermediate message types, they're just too much overhead at this stage. We'll look at adopting schema based formats for versioning etc in the future. For now we can just use bincode.
* Fully documents the public API and adds lints
* Adds traits and types for hashing types implementing serde traits
  * Adds [BCS](https://docs.rs/bcs/latest/bcs/) dependency for canonical serialization
* Adds traits and types for hash commitments of types implementing serde traits
* Removes old dead code

### DIFF
--- a/mpc/mpc-core/src/hash.rs
+++ b/mpc/mpc-core/src/hash.rs
@@ -11,6 +11,13 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Hash([u8; 32]);
 
+impl Hash {
+    /// Returns the hash as a byte slice
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
+}
+
 impl From<[u8; 32]> for Hash {
     fn from(bytes: [u8; 32]) -> Self {
         Self(bytes)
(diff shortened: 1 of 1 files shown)


mpc/mpc-core/src/commit.rs
```diff
@@ -0,0 +1,124 @@
...
+        Ok(())
+    }
+
+    /// Returns the data
+    pub fn data(&self) -> &T {
+        &self.data
+    }
+
+    /// Returns the data
+    pub fn to_inner(self) -> T {
```
th4s, line 68: `into_inner` probably fits better

mpc/mpc-core/src/commit.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Creates a new decommitment
+    pub fn new(data: T) -> Self {
+        Self {
+            nonce: Nonce::random(),
+            data,
+        }
+    }
+
+    /// Creates a hash commitment
+    pub fn commit(&self) -> Hash {
```
th4s, line 49: Can we rename this, or maybe the other one, as it collides with `HashCommit::commit`?

mpc/mpc-core/src/hash.rs
```diff
@@ -0,0 +1,85 @@
...
+/// # use serde::{Deserialize, Serialize};
+///
+/// #[derive(Serialize, Deserialize)]
+/// pub struct Foo(u64);
+///
+/// // All instances of `Foo` will be hashed with the domain separator "FOO"
+/// impl_domain_separated_hash!(Foo, "FOO");
+/// ```
+#[macro_export]
+macro_rules! impl_domain_separated_hash {
```
th4s, line 66: This could be `derive` macro, but no need to adapt. What is the advantage here of having a macro vs. a trait with a default implementation and some associated const?

th4s (commented):
gw, ack

mpc/mpc-core/src/commit.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Debug, thiserror::Error)]
+#[allow(missing_docs)]
+pub enum CommitmentError {
+    #[error("Invalid decommitment")]
+    InvalidDecommitment,
+}
+
+/// A randomly generated 32 byte nonce
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct Nonce([u8; 32]);
```
themighty1, line 17: nonces prevent replay attacks. this is not why we add randomness to the hash. `Salt` is a more fitting name here.

mpc/mpc-core/src/hash.rs
themighty1, line 66: Can we have a test that this macro works as expected?

themighty1 (approved):
ack with comments

PR #219, opened 2023-04-18
Refactor mpc-garble-core

sinui0:
This PR greatly simplifies our core garbled circuit types, building on top of the improvements to `mpc-circuits` brought in #212.

# Changes
* Completely documented the public API and added lints
* Serde all the things
* Rewrote the core types for representing encoded values, building on #212 
  * With these changes came immense simplifications to our types for encodings, decodings, label commitments, etc. Many things are now statically enforced by the type system, removing a crap load of boiler plate for data validation.
  * Encoded values are _decoupled_ from the old `GarbledCircuit` type (which no longer exists :pray: ). Values are independent of circuits.
  * The `Label` type is mostly hidden from the public API.
  * Decodings are now size optimal. Eg, the decoding for a `u8` is represented as a `u8`, whereas previously we were sending `Vec<bool>`
* The encoder is now idempotent. I understand the safety concerns for this, but this is a _core_ crate, not to be used be the general public. Structures for managing the state required to prevent duplicate encodings are coming in the `mpc-garble` crate in the following PR.
* Garbled circuit streaming :rocket: Now both the generator and evaluator can execute the circuit concurrently. This was not necessary but didn't take much and I was in the area anyhow.

Needs more tests

### DIFF
--- a/mpc/mpc-circuits/src/types.rs
+++ b/mpc/mpc-circuits/src/types.rs
@@ -34,7 +34,7 @@ pub trait ToBinaryRepr: ToBitsIter + Into<Value> {
     fn new_bin_repr(nodes: &[Node<Feed>]) -> Result<Self::Repr, TypeError>;
 }
 
-/// A type to which the value type can be statically determined.
+/// A type for which the value type can be statically determined.
 pub trait StaticValueType {
     /// The value type of the type.
     fn value_type() -> ValueType;


mpc/garble/mpc-garble-core/src/encoding/value.rs
```diff
@@ -0,0 +1,748 @@
...
+}
+
+define_encoded_variant!(Bit, bool, 1);
+define_encoded_variant!(U8, u8, 8);
+define_encoded_variant!(U16, u16, 16);
+define_encoded_variant!(U32, u32, 32);
+define_encoded_variant!(U64, u64, 64);
+define_encoded_variant!(U128, u128, 128);
+
+macro_rules! define_decoding {
```
th4s, line 386: Same for this macro.

mpc/garble/mpc-garble-core/src/encoding/value.rs
```diff
@@ -0,0 +1,748 @@
...
+                    h == pair[0] || h == pair[1]
+                }) {
+                    Ok(())
+                } else {
+                    Err(ValueError::InvalidCommitment)
+                }
+            }
+
+            // We use a truncated Blake3 hash to commit to the labels
+            fn compute_hash(block: Block) -> Block {
```
th4s, line 667: Can't we use `DomainSeparatedHash` here?

mpc/garble/mpc-garble-core/src/evaluator.rs
```diff
@@ -31,60 +51,152 @@ pub(crate) fn and_gate<C: BlockCipher<BlockSize = U16> + BlockEncrypt>(
     Label::new(w_g ^ w_e)
 }
 
-/// Evaluates half-gate garbled XOR gate
-#[inline]
-pub(crate) fn xor_gate(x: &Label, y: &Label) -> Label {
-    *x ^ *y
+/// Core evaluator type for evaluating a garbled circuit.
+pub struct Evaluator {
```
th4s, line 55: Can we add some description for the fields?

mpc/garble/mpc-garble-core/src/generator.rs
```diff
@@ -37,73 +55,182 @@ pub(crate) fn and_gate<C: BlockCipher<BlockSize = U16> + BlockEncrypt>(
...
+    /// Returns the hash of the encrypted gates.
+    pub fn hash(&self) -> Option<Hash> {
+        self.hasher.as_ref().map(|hasher| {
+            let hash: [u8; 32] = hasher.finalize().into();
+            Hash::from(hash)
+        })
+    }
+}
+
+impl Iterator for Generator {
```
th4s, line 157: I am not sure if it is a good idea to have state-modifying behavior inside an iterator implementation, which has nothing to do with iteration. You feed things into the hasher and modify `gid` while iterating.

In contrast, the evaluator does these things in `evaluate`. Maybe it is better to have a special method which does these things for the generator as well.

mpc/garble/mpc-garble-core/src/lib.rs
```diff
@@ -1,156 +1,129 @@
...
+pub use circuit::EncryptedGate;
+pub use encoding::{
+    state as label_state, ChaChaEncoder, Decoding, Delta, Encode, EncodedValue, Encoder,
+    EncodingCommitment, EqualityCheck, Label, ValueError,
 };
+pub use evaluator::{Evaluator, EvaluatorError};
+pub use generator::{Generator, GeneratorError};
+
+/// Fixed key used for AES encryption
+pub(crate) static CIPHER_FIXED_KEY: [u8; 16] = [69u8; 16];
```
th4s, line 23: Can we add a short comment, why it is ok to use a hard-coded encryption key?

th4s (changes requested):
Awesome :100:, nice PR! I have some comments :)

mpc/garble/mpc-garble-core/src/encoding/value.rs
```diff
@@ -0,0 +1,748 @@
...
+/// A trait for encoding values.
+pub trait Encode: ToBitsIter {
+    /// The encoded value type.
+    type Encoded;
+
+    /// Encodes a value using the provided delta and labels.
+    fn encode(delta: Delta, labels: &[Label]) -> Result<Self::Encoded, ValueError>;
+}
+
+macro_rules! define_encoded_value {
```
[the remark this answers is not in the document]
th4s: I am fine with both.

mpc/garble/mpc-garble-core/src/generator.rs
th4s: I think my thought was, that if the `Generator` is used without iteration somehow, then the hash is wrong.

mpc/garble/mpc-garble-core/src/encoding/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    /// Creates a new label
+    #[inline]
+    pub fn new(value: Block) -> Self {
+        Self(value)
+    }
+
+    /// Returns inner block
+    #[inline]
+    pub(crate) fn into_inner(self) -> Block {
```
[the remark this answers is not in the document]
th4s: I always have to look up the non-trivial case :D

mpc/garble/mpc-garble-core/src/encoding/value.rs
th4s: Ok then let's keep it.

mpc/garble/mpc-garble-core/src/generator.rs
th4s: Hmm I see. Ok then let's keep it and maybe add a note?

themighty1 (approved):
ACK, gw

th4s (approved):
:rocket: ack

PR #221, opened 2023-04-24
Deprecate dep attribute, add test macro

sinui0:
This PR deprecates the `dep` attribute in the `mpc-circuits-macros` crate in favor of having that functionality built in the `trace` macro itself. I also added a small helper macro for testing a circuit using a function with the same signature.


th4s (approved):
ack

PR #224, opened 2023-04-24
Implement free-xor for encoded values

sinui0:
This PR re-implements free-xor for encoded values. This functionality was lost during the rewrite, but it is useful.


th4s (approved):
:+1: ack

themighty1 (approved):
ack

PR #230, opened 2023-05-01
Refactor PRF for new MPC API

sinui0:
This PR refactors our MPC HMAC-SHA256 impl to use the new MPC API.

# Changes
- Lints + Docs + Clippy :paperclip: 
- Uses new `mpc-circuits` API to build the circuits. I avoided using the `trace` macro, as I encountered some issues with it. Will hold off on using it until we have more time to stabilize it.
- Uses new `mpc-garble` API, which simplified the crate quite a bit. There is no longer a `Leader/Follower` distinction and just the 1 `Prf` trait.
- Removed circuit pre-building. I believe these circuits come up to ~100Mb which is too large to ship with our binaries. Instead we can build them at runtime. Right now this is pretty inefficient but see tlsnotary/mpz#5 for a solution to our woes.


th4s (approved):
:rocket: ack

PR #236, opened 2023-05-04
Ringbuffer

th4s:
This PR adds a ringbuffer to our `utils-aio` crate. It supports `Read`, `Write`, `AsyncRead` and `AsyncWrite`. I intend to use this for what is currently called `ExchangeBuffer` in this overview (last picture)
https://excalidraw.com/#room=38aba876235c50f6d8db,RhSP5NmWaqzMpa6ffRXCxg

### DIFF
--- a/utils/utils-aio/src/lib.rs
+++ b/utils/utils-aio/src/lib.rs
@@ -8,6 +8,7 @@ pub mod factory;
 #[cfg(feature = "mux")]
 pub mod mux;
 pub mod non_blocking_backend;
+pub mod ring_buffer;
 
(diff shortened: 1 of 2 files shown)


th4s:
Oh I thought this would be a great opportunity to go a little bit deeper on async and atomics. I mainly did this because I imagined that we want to have something which on one side looks like the sender part of a channel and on the other side is a raw buffer. When I looked around there were not that many buffers supporting `AsyncRead/Write`. It did not occur to me that tokio supports channel adapters in its utils...

I think the implementation should be fine but I see your concern... Hm.....
I can give it a try with tokio and see if it works fine. If yes then I am fine with ditching this. In the meantime we can leave it open?

PR #240, opened 2023-05-11
Share Conversion Cleanup

sinui0:
This PR cleans up our share conversion crates (and updates GHASH).

# Changes
- Lints + Clippy :paperclip: 
- Deprecates our share-conversion actor, it is not necessary as discussed.
- Simplifies the `Field` trait to no longer have an `Encoding` associated type. This information was moved to the `BlockSerialize` trait in `mpc_core`.
- Adds OT supertraits to provide blanket impls for `Field` types.
- Modifies share conversion traits to receive `&self` instead of `&mut self` for multithread access, and wrapped internal state in a Mutex to facilitate this.
- Updated the `Gf2_128` type constructor to make it easier to reason about, and tweaked `Mul` to match the NIST implementation

In the future we will want to modify out `clmul` crate to support `clmul` with _and_ without modular reduction so that we can use the optimized + constant time implementations.

### DIFF
--- a/mpc/share-conversion/mpc-share-conversion-core/src/fields/gf2_128.rs
+++ b/mpc/share-conversion/mpc-share-conversion-core/src/fields/gf2_128.rs
@@ -16,7 +16,7 @@ impl Gf2_128 {
     /// Creates a new field element from a u128,
     /// mapping the integer to the corresponding polynomial
     ///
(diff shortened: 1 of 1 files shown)


th4s (approved):
Very nice refactor :rocket:

PR #242, opened 2023-05-14
Refactor share conversion, again.

sinui0:
This PR refactors our share conversion crates in a number of ways described below. Note that I avoided making alterations to the protocol algorithms with the exception of the `Mul` impl on `Gf2_128`.

# Changes
- Docs, lints + clippy :paperclip: 
- Renamed some methods for clarity, I hope you agree
- Deprecates the share conversion actor in favor of lightweight Arc + Mutex wrappers found in `mpc-share-conversion/src/converter.rs`
- Simplifies the `Field` trait to no longer have an `Encoding` associated type. This information was moved to the `BlockSerialize` trait in `mpc_core`.
- Adds OT supertraits to provide blanket impls for `Field` types.
- Modifies share conversion traits to receive `&self` instead of `&mut self`, and wrapped internal state in a Mutex to facilitate this.
  - *Note* Concurrent conversion is not actually supported at the moment. To support this we will have to change the way we record to the tape, using a unique ID approach similar to what we do in `mpc-garble`. Trying to use the converters concurrently right now will just result in a tape verification error if a race condition occurred (but not an actual security issue).
- Consolidated the `A2M` and `M2A` functionality into a single instance. Both protocols can be recorded on the same tape. This reduces a lot of boilerplate.
- Updated the `Gf2_128` type to make it easier to reason about, and tweaked `Mul` to match the NIST implementation.
- Updated dependent crates
- Implemented `serde` derives for the protocol messages

### DIFF
--- a/mpc/share-conversion/mpc-share-conversion/src/lib.rs
+++ b/mpc/share-conversion/mpc-share-conversion/src/lib.rs
@@ -1,5 +1,9 @@
-//! This crate provides a protocol for converting secret-shared finite field elements
+//! This crate provides a semi-honest protocol (with optional covert-security) for converting secret-shared finite field elements
 //! between additive and multiplicative representations.
+//!
+//! The protocol is based on `Two Party RSA Key Generation [Gil99]` which devised a method for
+//! converting additive shares of a finite field element into multiplicative shares (A2M). We use a similar technique
+//! to convert multiplicative shares into additive shares (M2A), inspired by `Efficient Secure Two-Party Exponentiation [YCCL11]`.
 
 #![deny(missing_docs, unreachable_pub, unused_must_use)]
 #![deny(clippy::all)]


mpc/share-conversion/mpc-share-conversion/src/converter.rs
```diff
@@ -0,0 +1,278 @@
...
+use crate::{
+    AdditiveToMultiplicative, GilboaReceiver, GilboaSender, MultiplicativeToAdditive,
+    OTReceiveElement, OTSendElement, ReceiverConfig, SenderConfig, ShareConversionChannel,
+    ShareConversionError,
+};
+
+/// The share conversion sender
+pub struct ConverterSender<F: Field, OT> {
+    ot: OT,
+    sender: Option<Arc<GilboaSender<F>>>,
```
th4s, line 15: If I remember correctly, only one version was called "Gilboa". I think it was M2A from this paper on page 6 https://link.springer.com/content/pdf/10.1007/3-540-48405-1_8.pdf

mpc/share-conversion/mpc-share-conversion/src/converter.rs
```diff
@@ -0,0 +1,278 @@
...
+                &self.ot,
+                &input.into_iter().map(Share::new_mul).collect::<Vec<_>>(),
+            )
+            .await
+            .map(|shares| shares.into_iter().map(|share| share.to_inner()).collect())
+    }
+}
+
+/// A handle to a sender
+pub struct ConverterSenderHandle<F: Field, OT> {
```
th4s, line 103: Why exactly do we need that?

th4s (approved):
Nice :rocket: You now chose an enum approach for `Add` and `Mul` shares. It's very interesting to see the difference considering implementation and API, since we had it generic before. What were your thoughts? I feel that the code gets simpler (which is definitely a big advantage). On the other hand we move a lot from compile-time typesystem checks to runtime. What was your trigger to do this :upside_down_face: ?

mpc/share-conversion/mpc-share-conversion-core/src/shares.rs
```diff
@@ -0,0 +1,273 @@
...
+        Self::Add(AddShare::new(share))
+    }
+
+    /// Create a new multiplicative share.
+    pub fn new_mul(share: T) -> Self {
+        Self::Mul(MulShare::new(share))
+    }
+
+    /// Returns the type of the share.
+    pub fn ty(&self) -> ShareType {
```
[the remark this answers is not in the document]
th4s: Ahh ok I did not know about this. I prefer `typ` but fine with the other one, too, if that is some convention.

th4s:
No concerns. I just wanted to know :) I think the API is nicer now.

mpc/share-conversion/mpc-share-conversion/src/sender.rs
```diff
@@ -0,0 +1,150 @@
...
+            for share in shares {
+                let (new_share, summands) = share.convert(&mut state.rng);
+                new_shares.push(new_share);
+                all_summands.extend(summands);
+            }
+
+            (new_shares, all_summands, ot_id)
+        };
+
+        // Send OT shares to the receiver and increment batch counter
```
themighty1, line 101: "and increment batch counter" is not accurate for this line

mpc/share-conversion/mpc-share-conversion/src/tape.rs
```diff
@@ -0,0 +1,95 @@
+use crate::TapeVerificationError;
+
+use mpc_share_conversion_core::{fields::Field, Share};
+use rand::SeedableRng;
+use rand_chacha::ChaCha12Rng;
+
+/// A tape which allows to record inputs and outputs of the conversion.
+/// The sender can reveal his tape (thus revealing all his secret inputs) to the receiver
+/// who will check it against her tape to detect any malicious behaviour of the sender.
```
themighty1, line 9: Can we make an extra note here that rng seed is also revealed by the sender, otherwise it may seems as if tape is all that is needed for the replay.

themighty1 (changes requested):

themighty1:
lgtm

PR #243, opened 2023-05-15
OT reveal trigger

sinui0:
This PR modifies our OT Sender actor to trigger the reveal of every OT at once, removing the "MarkForReveal" functionality.

Synchronization can be managed in the outer context, rather than putting the responsibility on dependent protocols to indicate when they are ready. With this approach the caller of `sender.reveal()` must know when its time to reveal everything.


th4s (approved):
ack

themighty1 (approved):
ack

PR #244, opened 2023-05-17
Mock OT typecasting

sinui0:
This PR removes the generic from our Mock OT and uses downcasting of `dyn Any` so that a mock pair can support transferring any value type.


th4s (approved):
ack

themighty1 (approved):
ack

PR #245, opened 2023-05-18
PeerEncodings trait

themighty1:
This PR
adds a PeerEncodings trait to DEAP VM to enable to extract peer's encodings. These encodings can be used by the Prover to create commitments.


th4s (approved):
nice :rocket:, only some minor things to please clippy, since we use `#![deny(clippy::all)]`.

PR #247, opened 2023-05-22
Consolidate KE trait

sinui0:
This PR consolidates the key exchange traits into a single trait. We already use dynamic dispatch in the implementation, there is no benefit with having multiple interfaces and it has resulted in duplication downstream.

This PR also hides away implementation details related to communicating the keys, and drops the `private_key` from memory after it is used.


th4s (approved):
Nice :rocket:

PR #249, opened 2023-05-23
Simplified Muxer

sinui0:
This PR deprecates our actor based muxer in favor of a simpler wrapper.

# Changes
- New `uid-mux` crate which is a simple wrapper around yamux for adding the unique stream id functionality we want.
- Deprecates `actor-mux`
- Modifies mux traits to take id by reference.


th4s (approved):
ack :rocket:

themighty1 (commented):
ack

PR #250, opened 2023-05-23
IO trait refactor

sinui0:
This PR does some refactoring of our IO + Mux traits. I discovered that we missed something which had downstream effects virtually everywhere we do IO: All our trait bounds for streams were: `Stream<Item = T>` which implies that the channel is infallible :man_facepalming:  I've gone and updated everything to `Stream<Item = Result<T, std::io::Error>>`.

I also cleaned up our mux traits a bit and wrote a wrapper which applies a bincode codec to any duplex bytestream returned by our muxer. It provides blanket impls for serde types so should be pretty ergonomic.

Lastly, I spruced up the `expect_msg_or_err` macro to be a little bit more ergonomic, it now just uses `std::io::Error` for unexpected messages.

### DIFF
--- a/aead/src/aes_gcm/mod.rs
+++ b/aead/src/aes_gcm/mod.rs
@@ -119,8 +119,7 @@ impl MpcAesGcm {
             }
             Role::Follower => {
                 // Wait for commitment from leader
-                let commitment =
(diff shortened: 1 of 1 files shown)


th4s (approved):
nice, ack

themighty1 (commented):
gw, ack

PR #252, opened 2023-05-23
TLS-MPC minimal functionality

sinui0:
This PR provides the minimal functionality for our TLS-MPC implementation.

It implements both roles and the backend trait to integrate with the TLS client.


th4s (approved):
:rocket: really cool to see this come together, ack

PR #254, opened 2023-05-26
Updated rs_merkle to 1.4.1

heeckhau:
re #246


themighty1 (approved):
Thanks, @heeckhau , looks good to merge.
`rs_merkle` does not return an error variant for invalid proofs but only `false` (same `false` as when verification fails), so there's nothing else that we can do here.

PR #256, opened 2023-05-29
Refactor tls(n)

sinui0:
This PR refactors `tls-backend` and `tlsn-core` to use more existing types from `rustls`.

# Changes
- Lints + Clippy for `tlsn-core`
- Moved some modules from `tls-client` to `tls-core`
- Added functionality for forwarding certificate and key exchange details to the `tls-backend`
- Added `serde` derives to `tls-core`
- Cleaned up `tlsn-core` modules and replaced some types with existing types from `tls-core` such as `WebPkiVerifier` for using `webpki` for certificate and signature validation. I don't expect there was anything wrong with the types introduced by @themighty1 but I reckon it is better to reuse as much from rustls as possible.

### DIFF
--- a/tlsn/tlsn-core/tests/api.rs
+++ b/tlsn/tlsn-core/tests/api.rs
@@ -166,6 +166,16 @@ fn test_api() {
     let SignedSessionHeader { header, signature } = bincode::deserialize(&msg_bytes).unwrap();
     //---------------------------------------
 
+    // User verifies the signature
+    #[allow(irrefutable_let_patterns)]
+    if let Signature::P256(signature) = signature {
+        notary_pubkey
+            .verify(&header.to_bytes(), &signature)
+            .unwrap();
+    } else {
+        panic!("Notary signature is not P256");
+    };
+
     // User verifies the header and stores it with the signature in NotarizedSession
     header.check_artifacts(&artifacts).unwrap();
 


th4s (approved):
ack

tlsn/tlsn-core/tests/api.rs
```diff
@@ -0,0 +1,232 @@
...
+        header,
+        signature: signature.into(),
+    };
+
+    //---------------------------------------
+    let msg_bytes = bincode::serialize(&msg).unwrap();
+    let SignedSessionHeader { header, signature } = bincode::deserialize(&msg_bytes).unwrap();
+    //---------------------------------------
+
+    // User verifies the header and stores it with the signature in NotarizedSession
```
themighty1, line 169: I cant see the step where the User verifies the Notary's signature. I had it in my original test.

themighty1 (commented):
lgtm, except a check is missing
also, is it possible to mark those files/types which were borrowed from rustls as such to help the reviewer concentrate only on the code which we wrote.

tlsn/tlsn-core/tests/api.rs
themighty1: One reason is consistency of the approach: we chose the approach of the User validating the data at the time when it is stored. 
Also if a malicious Notary provides the wrong sig and the User doesn't detect the wrong sig immediately, it may disrupt the User who counts on being able to create a proof later.

themighty1:
> > is it possible to mark those files/types which were borrowed from rustls as such to help the reviewer concentrate only on the code which we wrote
> 
> I'm not quite sure what you have in mind with this. We can indicate that _most_ things in `tls-core` are borrowed from `rustls` but trying to maintain a clean record of every component is likely to become quite burdensome

ok, i'll take a closer look to see what approach makes sense.

PR #257, opened 2023-05-30
Async client io

sinui0:
This PR finishes providing async support for the `tls-client`.

Changes:
- Added `tls-client-async` which provides a complete wrapper around the `ClientConnection` for handling it's lifecycle and exposing a simple `AsyncWrite + AsyncRead` interface for the user.
- Added `tls-server-fixture` which provides a fixture for testing against a server using a self signed CA+EE cert. I copied a simple example from `hyper` to run an echo server.
- Updated `tls-mpc` test to use the server fixture and client wrapper

### DIFF
--- a/integration-tests/tests/test.rs
+++ b/integration-tests/tests/test.rs
@@ -23,7 +23,7 @@ use utils_aio::{codec::BincodeMux, mux::MuxChannel};
 async fn test() {
     let mut rng = ChaCha20Rng::seed_from_u64(0);
 
-    let (leader_socket, follower_socket) = tokio::io::duplex(2 << 25);
+    let (leader_socket, follower_socket) = tokio::io::duplex(1 << 25);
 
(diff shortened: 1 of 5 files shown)


tls/tls-client-async/src/conn.rs
```diff
@@ -0,0 +1,116 @@
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<(), std::io::Error>> {
+        Pin::new(&mut self.tx_sender).poll_flush(cx)
+    }
+
+    fn poll_close(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<(), std::io::Error>> {
```
th4s, line 100: Shouldn't we also call `poll_close` on `self.tx_sender`?

themighty1 (approved):
gw!

PR #258, opened 2023-05-31
fix deadlock, removing tokio codec

sinui0:
This PR fixes a deadlock issue in `uid-mux`.

The current implementation uses `LengthDelimitedCodec` for transmitting the unique `stream_id` to the mux peer. The issue is that `LengthDelimitedCodec` performs internal buffering which is dropped when the stream is pulled back out of the codec before returning it to the caller. This data loss causes a deadlock.

I've replaced the codec with a simple manual impl of length delimited framing for the `stream_id`.


themighty1 (commented):
ack

PR #262, opened 2023-06-06
Release preparation for crate `point-addition`

th4s:
- Improve manifest
- Check for panics/unwraps
- Add tracing
- Improve doc

### DIFF
--- a/point-addition/src/conversion.rs
+++ b/point-addition/src/conversion.rs
@@ -68,7 +68,7 @@ where
     /// Perform the conversion of P = A + B => P_x = a + b
     ///
     /// Since we are only interested in the x-coordinate of P (for the PMS) and because elliptic
-    /// curve point addition is an expensive operation in 2PC, we secretly share the x-coordinate
+    /// curve point addition is an expensive operation in 2PC, we secret-share the x-coordinate
     /// of P as a simple addition of field elements between the two parties. So we go from an EC
     /// point addition to an addition of field elements for the x-coordinate.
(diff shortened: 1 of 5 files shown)


point-addition/src/conversion.rs
```diff
@@ -65,7 +71,12 @@ where
     /// curve point addition is an expensive operation in 2PC, we secretly share the x-coordinate
```
themighty1, line 71: nit: it is more common to say "we secret-share" vs "we secretly share"

point-addition/src/lib.rs
```diff
@@ -1,15 +1,26 @@
+//! A library for converting two-party shares of an elliptic curve (EC) point's x-coordinate under
```
themighty1, line 1: iiuc, the 2 parties each start with an EC point. These 2 EC points are additive shares of another EC point. So, I took a stab at simplifying this description:

A secure two-party computation (2PC) library for converting additive shares of an elliptic curve (EC) point into additive shares of said point's x-coordinate. <Optionally an extra sentence mentioning that the x-coordinate is a field element>

point-addition/src/lib.rs
```diff
@@ -18,11 +29,15 @@ pub enum PointAdditionError {
     Coordinates,
 }
 
+/// A trait for secretly sharing the sum of two elliptic curve points as a sum of field elements
```
themighty1, line 32: secret-sharing

themighty1 (approved):
ack

PR #272, opened 2023-06-15
Release preparation for crate `key-exchange`

th4s:
- Tidy up manifest
- Improve comments
- Add tracing

### DIFF
--- a/components/key-exchange/src/exchange.rs
+++ b/components/key-exchange/src/exchange.rs
@@ -3,11 +3,11 @@
 use async_trait::async_trait;
 use futures::{SinkExt, StreamExt};
 use mpz_garble::{Decode, Execute, Memory};
-use std::borrow::Borrow;
 
 use mpz_share_conversion_core::fields::{p256::P256, Field};
 use p256::{EncodedPoint, PublicKey, SecretKey};
 use point_addition::PointAddition;
+use std::fmt::Debug;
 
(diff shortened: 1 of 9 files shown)


key-exchange/Cargo.toml
```diff
@@ -1,5 +1,10 @@
 [package]
 name = "tlsn-key-exchange"
+authors = ["TLSNotary Team"]
+description = "Implementation of the TLSN-specific key-exchange protocol"
```
themighty1, line 4: Let's use the phrase "TLSNotary-specific" since the less ambiguity in the public description, the better.

themighty1 (approved):

components/key-exchange/src/exchange.rs
```diff
@@ -50,6 +69,10 @@ where
     /// * `point_addition_receiver` - The point addition receiver instance used during key exchange
     /// * `executor`                - The MPC executor
     /// * `config`                  - The config used for the key exchange protocol
+    #[cfg_attr(
+        feature = "tracing",
+        tracing::instrument(level = "info", skip(channel, executor), ret)
```
themighty1, line 74: wu must also skip `point_addition_sender` and `point_addition_receiver` since the contain points from which session keys can be derived

themighty1 (changes requested):
ack, one change needed

PR #274, opened 2023-06-19
Release preparation for workspace `prf`

th4s:
- Tidy up manifests
- Improve doc
- Check for panics
- Add tracing

### DIFF
--- a/components/prf/hmac-sha256/src/prf.rs
+++ b/components/prf/hmac-sha256/src/prf.rs
@@ -121,7 +121,7 @@ where
 
     #[cfg_attr(
         feature = "tracing",
-        tracing::instrument(level = "debug", skip(self), err)
+        tracing::instrument(level = "debug", skip(self, label), err)
     )]
     async fn internal_compute_vd(
         &mut self,
--- a/components/prf/hmac-sha256/src/prf.rs
+++ b/components/prf/hmac-sha256/src/prf.rs
@@ -14,7 +14,7 @@ where
     executor: E,
 }
 
-/// Internal state of the PRF.
+/// Internal state of [MpcPrf].
 #[derive(Debug, Clone)]
 #[allow(missing_docs)]
 pub enum State {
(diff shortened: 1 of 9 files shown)


prf/hmac-sha256-circuits/src/hmac_sha256.rs
```diff
@@ -87,7 +89,8 @@ pub fn hmac_sha256_partial(key: &[u8]) -> ([u32; 8], [u32; 8]) {
 ///
 /// * `outer_state` - 256-bit outer state
 /// * `inner_state` - 256-bit inner state
-/// * `msg` - N-byte message
+/// * `msg`         - N-byte message
+#[cfg_attr(feature = "tracing", tracing::instrument(level = "trace"))]
```
themighty1, line 93: Im sure we need to NOT include some of the args into the trace.

themighty1 (changes requested):
A general comment:
we don't want to include in the trace any material which would allow a malicious Notary who gets hold of the debug log to construct the full session keys and to decrypt the whole TLS transcript. 

I would suggest that we `skip` absolutely everything (that includes also skipping `inner_state`, `outer_state`). This will make it simpler to reason about: "we just dont log anything from the PRF, so it's impossible for the Notary who gets hold of the logs to glean anything".

Otherwise, we (and also future code auditors) would have to carefully examine whether or not the specific material which we log creates an issue.

th4s:
I adapted the logging. Now we only log `client_random`, `server_random` and `handshake_hash`.

components/prf/hmac-sha256/src/prf.rs
```diff
@@ -152,6 +178,10 @@ where
             .await
     }
 
+    #[cfg_attr(
+        feature = "tracing",
+        tracing::instrument(level = "debug", skip(self, handshake_hash), err)
```
themighty1, line 183: for consistency, let's **NOT** skip `handshake_hash` here since we are not skipping it elsewhere

themighty1 (approved):
gw!

PR #281, opened 2023-06-21
Reorganizing sources: move tlsn to toplevel, components to components

heeckhau:


heeckhau:
@sinui0 this PR moves the different components to a `components` subfolder.

I did not move the tlsn members to the toplevel. I tried that, but this resulted in problems inside the components folder, because the `Cargo.toml` file of tlsn was than treated as parent configuration file.
I guess we will only be able to do this, once we can combine everything in one workspace.

heeckhau:
The problem is in the other direction. The components can no longer be build if the tlsn `Cargo.toml` file appears in the top folder.

```
error: failed to load manifest for workspace member `/Users/heeckhau/tlsnotary/tlsn/components/tls/tls-mpc`

Caused by:
  failed to load manifest for dependency `tlsn-block-cipher`

Caused by:
  failed to parse manifest at `/Users/heeckhau/tlsnotary/tlsn/components/cipher/block-cipher/Cargo.toml`

Caused by:
  error inheriting `async-trait` from workspace root manifest's `workspace.dependencies.async-trait`

Caused by:
  `dependency.async-trait` was not found in `workspace.dependencies`
```

PR #286, opened 2023-07-03
Release preparation for workspace `tlsn`

th4s:
- add tracing
- tidy up manifests
- improve doc
- repair integration test

### DIFF
--- a/tlsn/tlsn-core/src/lib.rs
+++ b/tlsn/tlsn-core/src/lib.rs
@@ -1,13 +1,13 @@
 //! This crate contains types used by the Prover, the Notary, and the Verifier
 
-// TODO: #![deny(missing_docs)]
-#![deny(unreachable_pub, unused_must_use)]
+#![deny(missing_docs, unreachable_pub, unused_must_use)]
 #![deny(clippy::all)]
 #![forbid(unsafe_code)]
 
 pub mod commitment;
 mod error;
 #[cfg(any(test, feature = "fixtures"))]
+#[allow(missing_docs)]
 pub mod fixtures;
 mod handshake_summary;
 pub(crate) mod inclusion_proof;
@@ -37,6 +37,9 @@ pub use transcript::{Direction, Transcript, TranscriptSlice};
 /// This value must not exceed bcs's MAX_SEQUENCE_LENGTH limit (which is (1 << 31) - 1 by default)
 const MAX_TOTAL_COMMITTED_DATA: u64 = 1_000_000_000;
 
+/// The encoding id
+///
+/// A 64 bit Blake3 hash which is used for the garbling labels
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd, Hash)]
 pub struct EncodingId(u64);
 
--- a/tlsn/tlsn-core/src/lib.rs
+++ b/tlsn/tlsn-core/src/lib.rs
@@ -31,7 +31,7 @@ pub use substrings::{
 pub use transcript::{Direction, Transcript, TranscriptSlice};
 
 /// The maximum allowed total bytelength of all committed data. Used to prevent DoS during verification.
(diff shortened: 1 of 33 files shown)


tlsn/tlsn-core/src/lib.rs
```diff
@@ -37,6 +37,9 @@ pub use transcript::{Direction, Transcript, TranscriptSlice};
 /// This value must not exceed bcs's MAX_SEQUENCE_LENGTH limit (which is (1 << 31) - 1 by default)
 const MAX_TOTAL_COMMITTED_DATA: u64 = 1_000_000_000;
 
+/// The encoding id
+///
+/// A 64 bit Blake3 hash which is used for the garbling labels
```
themighty1, line 42: We agreed to avoid as much as possible any GC terminology in the `tlsn` crate. We need to re-word  "labels" and "garbling" with "plaintext encodings".

tlsn/tlsn-core/src/session/artifacts.rs
```diff
@@ -3,21 +3,22 @@ use mpz_core::commit::Decommitment;
 use crate::merkle::MerkleTree;
 use tls_core::{handshake::HandshakeData, key::PublicKey};
 
-// Various artifacts which the User holds at the end of the notarization session
+/// Various artifacts which the User holds at the end of the notarization session
```
themighty1, line 6: Let's switch all mentions of "User" to "Prover" to make things uniform.

tlsn/tlsn-core/src/session/mod.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
     Commitment, Direction, InclusionProof, SubstringsCommitment, SubstringsCommitmentSet,
 };
 
+/// A potentially notarized session
```
themighty1, line 24: Only prover deals with `NotarizedSession`. From his pov, it is not "potentially" correct, it is definitely correct. Maybe better "A validated notarized session stored by the Prover"

tlsn/tlsn-core/src/session/proof.rs
```diff
@@ -6,14 +6,17 @@ use tls_core::handshake::HandshakeData;
 use super::SessionHeader;
 use crate::signature::Signature;
 
+/// A proof for a TLS session
```
themighty1, line 9: lets s/TLS session/ TLSNotary session/ since it contains non-TLS stuff

tlsn/tlsn-core/src/transcript.rs
```diff
@@ -75,21 +80,25 @@ impl TranscriptSlice {
...
 
+    /// Gets a reference to data
     pub fn data(&self) -> &Vec<u8> {
         &self.data
     }
 }
 
-#[derive(Serialize, Deserialize, Clone, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 /// A [Transcript] consists of a stream of bytes which were sent to the server
```
themighty1, line 95: We changed Transcript to contain bytes only in one direction, so let's change this comment to:

/// A [Transcript] contains either a stream of bytes which were sent to the server
/// or a stream of bytes which were received from the server. The Prover creates
/// separate commitments to bytes in each direction.

tlsn/tlsn-notary/src/lib.rs
```diff
@@ -88,6 +100,7 @@ where
         Self { config, mux }
     }
 
+    /// Runs the notary server
```
themighty1, line 103: let's s/server/instance since this will perform only 1 notarization and then exit

themighty1 (approved):
ack

tlsn/tlsn-core/src/session/artifacts.rs
```diff
@@ -34,22 +35,27 @@ impl SessionArtifacts {
         }
     }
 
+    /// Getter for time
```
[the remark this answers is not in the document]
th4s: Agree, I now see that the struct's fields are private

PR #294, opened 2023-08-07
Async tls client fix. New server fixture and tests. Don't rely on close_notify.

themighty1:
This PR:
- introduces fixes in tls-client-async and tests.
- adds a new tls server fixture (with the existing hyper-based one it was not possible to test certain server behaviour).
-  does not treat absence of close_notify as an error.


themighty1:
thanks, I agree that it is better to handle the timeout in the outer context, I removed it.

PR #295, opened 2023-08-08
Finish removing Ring dependency

sinui0:
This PR finishes removing `ring` from `tls-core` and `tls-client` to support wasm builds. This was relatively straight forward as we only support P256 at the moment. I've verified locally that the `tls-client` crate now builds for `wasm32-unknown-unknown`. We will want to add this to our CI.

# Changes
- Replace usage of `ring` with functional equivalents from RustCrypto (key exchange, signing and hashing).
- Replace usage of `ring` constant time slice comparison with `subtle`
- Remove `Edd25519` and `Rsa` signing impls
- Remove some dead code


components/tls/tls-core/src/x509.rs
```diff
@@ -1,6 +1,37 @@
 // Additional x509/asn1 functions to those provided in webpki/ring.
 
-use ring::io::der;
+pub const CONSTRUCTED: u8 = 1 << 5;
+pub const CONTEXT_SPECIFIC: u8 = 2 << 6;
```
themighty1, line 4: did you mean 1 << 6 ?

themighty1 (approved):

Issue #296, opened 2023-08-08
Integration test is slow


heeckhau:
The integration test takes ~3 minutes atm: https://github.com/tlsnotary/tlsn/actions/runs/5794224056/job/15703397068#step:8:36 

Can we speed this up with running the test with `--release`?

PR #298, opened 2023-08-08
Test tlsn package with `--release` flag

heeckhau:
#296


themighty1 (approved):
ack on the idea of running the integration test with "--release"

PR #306, opened 2023-08-22
Fix range end checks, range end can be the same with data length because the end bracket is open

chcharcharlie:


chcharcharlie:
Should be straightforward but can provide more info if needed.

Found this when testing generating a notarized session and substring proofs, also when validating substring proofs of the entire received data.

themighty1 (approved):
Good catch, thanks for the PR

PR #307, opened 2023-08-23
Add quick_start.md

themighty1:
This PR adds a copy of `quick_start.md` from docs-mdbook/src/developers to the example folder for convenience (because README.md refers to it)
 
Also changes the URI the `simple_prover` uses to "/"
Before this change `simple_verifier` would output
`GET https://example.com/ HTTP/1.1`
now it outputs
`GET / HTTP/1.1`
which is the format a developer is more likely to expect and will cause less confusion


heeckhau (approved):
ack

PR #320, opened 2023-09-07
refactor: factor out generic from prover and notary

sinui0:
This PR simply factors out the generic over multiplexer in the prover and notary, and impls opaque_debug on state while I was in there.

This couples them to both `bincode` and `yamux`, but this is fine for now. If we want to support other formats and/or muxers we can add that as a feature later.


th4s (approved):
:+1:

PR #322, opened 2023-09-07
refactor: use Bytes in Transcript

sinui0:
This PR replaces the internal container for the transcript data with `Bytes` so that we can efficiently operate on slices of it without copying.


th4s (approved):
:+1:

PR #326, opened 2023-09-09
feat: PRF setup prior to handshake

### LINKED ISSUE #324: Pre-build circuits
The circuit builder API is not optimized yet, so building the TLS PRF circuits takes up to a few seconds. This slows things down quite a bit, and happens during the TLS connection.

Put the circuits into `OnceCell`s and build them at start up.

sinui0:
This PR adds a `setup` method to the PRF so that the circuits can be built prior to connecting to the application server. The circuits are built in parallel and no longer cause blocking in the operation of the PRF.

Additionally, the circuits are stored in static `OnceLock`, this will mitigate having to build them multiple times if the same process performs multiple sessions (Notary).

This should shave a couple seconds off the handshake time.

Closes #324


th4s (approved):
ack :+1:

themighty1:
ack

PR #327, opened 2023-09-10
refactor: decouple notarization

sinui0:
This PR decouples the Prover from the notarization flow, where we will support branching to either notarize or immediately to selective disclosure with an application specific verifier.


th4s (approved):
:+1:

themighty1:
ack, except Im not sure if the term "notarization" is used correctly.

PR #332, opened 2023-09-13
[WIP] refactor: use RangeSet

sinui0:
This PR refactors internal handling of ranges to use `RangeSet`.

It also switches to using `Range<usize>` instead of `Range<u32>`, which may be a point of discussion. I believe the initial concern was related to 32-bit arch handling when `range.end > u32::MAX`. However, I don't see this as an issue as the only case that  could happen is during deserialization which would return an error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021 . I think it is safe to say we won't be dealing with transcripts longer than `u32::MAX`. We almost exclusively use these ranges for indexing, so there is an unnecessary burden converting them everywhere.

### DIFF
--- a/tlsn/Cargo.toml
+++ b/tlsn/Cargo.toml
@@ -20,7 +20,7 @@ tlsn-tls-client-async = { path = "../components/tls/tls-client-async" }
 tls-server-fixture = { path = "../components/tls/tls-server-fixture" }
 uid-mux = { path = "../components/uid-mux" }
 
-tlsn-utils = { git = "https://github.com/tlsnotary/tlsn-utils", branch = "feature/range-utils" }
+tlsn-utils = { git = "https://github.com/tlsnotary/tlsn-utils", rev = "c42405c" }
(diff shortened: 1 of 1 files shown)


tlsn/tlsn-core/src/substrings/commitment.rs
```diff
@@ -43,13 +42,11 @@ impl SubstringsCommitmentSet {
         }
 
         // grand total in all of the commitments' ranges must be sane
-        let mut total_committed = 0u64;
+        let mut total_committed = 0;
         for commitment in &self.0 {
-            for r in commitment.ranges() {
-                total_committed += (r.end - r.start) as u64;
```
themighty1, line 49: earlier `r.end` and `r.start` were u32, so `total_committed` which was u64 could not be overflown. With the change to usize it can be overflown.

themighty1 (commented):
seems like it should be ok to switch to usize except in one place where i commented.
pls ping me for review again.

PR #338, opened 2023-09-19
feat: add add_pem() to RootCertStore

themighty1:
This PR adds a `add_pem()` method to `RootCertStore` which should make it easy to instantiate the prover with a private CA by calling 
```rust
let store = RootCertStore::empty().add_pem(<pem string>)?;
ProverConfig::builder()
    .root_cert_store(store)
    ...
```


th4s (approved):
nice :+1:

PR #341, opened 2023-09-20
refactor: simplify verifier flow

sinui0:
This PR simplifies the API for non-interactive verifiers. It encapsulates some functionality into `tlsn-core` rather than pushing it out to another crate to handle.

We should consider replacing the `Option<Signature>` in `SessionProof` and `SessionData` with just `Signature`. It seems there is no longer any reason why these types would exist without a signature from the Notary. The direct Prover-Verifier config will not  use these types.

# Changes
- Did some reorganizing of types/modules in `tlsn-core`
- Encapsulated verification of `SessionProof`
- Introduced `TlsProof` which is simply a wrapper of both a `SessionProof` and `SubstringsProof`.
- `SessionProof` now also contains the `ServerName`

### DIFF
--- a/tlsn/tlsn-core/src/proof/tls.rs
+++ b/tlsn/tlsn-core/src/proof/tls.rs
@@ -20,7 +20,7 @@ use crate::{
 /// Proof that a transcript of communications took place between a Prover and Server.
 #[derive(Debug, Serialize, Deserialize)]
 pub struct TlsProof {
-    /// Proof of the TLS handshake, server identity, and commitments to the the transcript.
+    /// Proof of the TLS handshake, server identity, and commitments to the transcript.
     pub session: SessionProof,
     /// Proof regarding the contents of the transcript.
     pub substrings: SubstringsProof,


tlsn/tlsn-core/src/proof/tls.rs
```diff
@@ -0,0 +1,126 @@
...
+    proof::SubstringsProof,
+    session::SessionHeader,
+    signature::{Signature, SignatureVerifyError},
+    NotaryPublicKey, ServerName,
+};
+
+/// Proof that a transcript of communications took place between a Prover and Server.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TlsProof {
+    /// Proof of the TLS handshake, server identity, and commitments to the the transcript.
```
th4s, line 23: ... and commitments to ~~the~~ the transcript.

th4s (approved):
Nice :+1:

Considering the `Option<Signature>`: Do you just want to more clearly separate these verifier flows or what is the motivation behind?

themighty1 (approved):

PR #343, opened 2023-09-21
fix: sort merkle indices

sinui0:
This PR fixes an issue in `tlsn-core` related to the merkle tree module expecting the leaf indices to be sorted, but they're not. Now they are.

### DIFF
--- a/tlsn/tlsn-core/src/merkle.rs
+++ b/tlsn/tlsn-core/src/merkle.rs
@@ -58,18 +58,27 @@ pub struct MerkleProof {
 
 impl MerkleProof {
     /// Checks if indices, hashes and leaves count are valid for the provided root
+    ///
+    /// # Panics
+    ///
+    /// - If the length of `leaf_indices` and `leaf_hashes` does not match.
+    /// - If `leaf_indices` contains duplicates.
     pub fn verify(
         &self,
         root: &MerkleRoot,
         leaf_indices: &[usize],
         leaf_hashes: &[Hash],
     ) -> Result<(), MerkleError> {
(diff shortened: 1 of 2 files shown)


th4s (approved):
:+1:

tlsn/tlsn-core/src/merkle.rs
```diff
@@ -144,7 +153,17 @@ impl MerkleTree {
     }
 
     /// Creates an inclusion proof for the given `indices`
+    ///
+    /// # Panics
+    ///
+    /// - if `indices` is not sorted.
```
themighty1, line 159: @sinui0 , instead of imposing the burden on the caller to sort the indices, we can just sort them right in this fn.

tlsn/tlsn-core/src/merkle.rs
```diff
@@ -211,59 +230,83 @@ mod test {
...
             proof
-                .verify(&tree.root(), &[2, 4, 3], &[leaf1, leaf4, leaf3])
+                .verify(&tree.root(), &[2, 3, 4], &[leaf1, leaf3, leaf4])
                 .err()
                 .unwrap(),
             MerkleError::MerkleProofVerificationFailed
         );
+    }
 
-        // fail because of the extra leaf which was not covered by the proof
```
themighty1, line 241: @sinui0 , I'm not finding this test among the new tests.

themighty1 (commented):
@sinui0, some changes may be necessary

themighty1:
@sinui0 , I wasn't sure if this was the intended behaviour, so I thought I'd ask. Ok, i'll open an issue.

PR #359, opened 2023-10-08
tlsn-core fixtures

sinui0:
This PR moves the fixtures (again) into `tlsn-core` and eliminates a cyclic dependency that currently exists between `tlsn-core` and `tlsn-fixtures`.

I would prefer to have the fixtures in a separate crate but we need to use them for tests in `tlsn-core` so this is path of least resistance without a bunch of duplication.

# Changes
- Moved `tlsn-fixtures` into a module in `tlsn-core`
- Added a feature flag `fixtures` to `tlsn-core`
- Added running tests with `--all-features` in our CI
- Added a suite of functions for constructing fixtures.

### DIFF
--- a/tlsn/tlsn-core/src/fixtures/cert.rs
+++ b/tlsn/tlsn-core/src/fixtures/cert.rs
@@ -1,12 +1,10 @@
 use tls_core::{
-    anchors::{OwnedTrustAnchor, RootCertStore},
     key::{Certificate, PublicKey},
     msgs::{
         codec::Codec,
         enums::{NamedGroup, SignatureScheme},
         handshake::{DigitallySignedStruct, Random, ServerECDHParams},
     },
-    verify::ServerCertVerifier,
 };
 
 use hex::FromHex;


tlsn/tlsn-core/src/fixtures/cert.rs
```diff
@@ -0,0 +1,369 @@
...
+    #[fixture]
+    fn cert_verifier() -> impl ServerCertVerifier {
+        super::cert_verifier()
+    }
+
+    /// Expect chain verification to succeed
+    #[rstest]
+    #[case(tlsnotary())]
+    #[case(appliedzkp())]
+    fn test_verify_cert_chain_sucess_ca_implicit(
```
themighty1, line 165:
```suggestion
    fn test_verify_cert_chain_success_ca_implicit(
```

tlsn/tlsn-core/src/fixtures/cert.rs
```diff
@@ -0,0 +1,369 @@
...
+                &mut std::iter::empty(),
+                &[],
+                SystemTime::UNIX_EPOCH + Duration::from_secs(data.time),
+            )
+            .is_err());
+    }
+
+    /// Expect to fail because the host name is not a valid DNS name
+    #[rstest]
+    fn test_check_dns_name_present_in_cert_fail_invalid_dns_name() {
```
themighty1, line 366: seems like this test should go to components/tls/tls-core/src/dns.rs

themighty1 (approved):
gw, some minor comments

th4s (approved):
:+1:

tlsn/tlsn-core/src/fixtures/cert.rs
themighty1: ok

PR #360, opened 2023-10-11
refactor: tlsn-verifier

sinui0:
This PR renames `tlsn-notary` to `tlsn-verifier` and encapsulates the notary specific functionality.

This sets us up to support IZK for selective disclosure by branching to a different state instead of notarizing.

# Changes
- Rename to `tlsn-verifier`
- Move verifier code into `tls` module
- Introduce type states and decouple notary specific functionality from the rest of TLS verification.
- Properly close the muxer/connection at the end

### DIFF
--- a/tlsn/tlsn-prover/src/tls/mod.rs
+++ b/tlsn/tlsn-prover/src/tls/mod.rs
@@ -286,7 +286,7 @@ impl Prover<state::Notarize> {
 
         let merkle_root = session_data.commitments().merkle_root();
 
-        let notarize_fut = async move {
+        let mut notarize_fut = Box::pin(async move {
             let mut channel = notary_mux.get_channel("notarize").await?;
 
             channel
@@ -309,7 +309,7 @@ impl Prover<state::Notarize> {
             let signed_header = expect_msg_or_err!(channel, TlsnMessage::SignedSessionHeader)?;
 
             Ok::<_, ProverError>((notary_encoder_seed, signed_header))
-        }
+        })
         .fuse();
 
         let (notary_encoder_seed, SignedSessionHeader { header, signature }) = futures::select_biased! {


tlsn/tlsn-verifier/src/tls/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+                ot_recv,
+                ot_fut,
+                gf2,
+                encoder_seed,
+            },
+        })
+    }
+
+    /// Runs the TLS verifier to completion, notarizing the TLS session.
+    pub async fn notarize<S: AsyncWrite + AsyncRead + Send + Unpin + 'static, T>(
```
th4s, line 112: Can we rename this function into `execute_full_notarization`, or something better and shorter :) which captures the fact that it just runs the whole notary workflow?

th4s (approved):
:+1:

tlsn/tlsn-verifier/src/tls/mod.rs
th4s: Maybe `setup_and_notarize`?

th4s: Because there is another `notarize` function, and I think this would communicate the intent more precisely.

themighty1:
still reviewing ...

tlsn/tlsn-prover/src/tls/mod.rs
```diff
@@ -306,9 +306,10 @@ impl Prover<state::Notarize> {
             let signed_header = expect_msg_or_err!(channel, TlsnMessage::SignedSessionHeader)?;
 
             Ok::<_, ProverError>((notary_encoder_seed, signed_header))
-        };
+        }
+        .fuse();
 
-        let (notary_encoder_seed, SignedSessionHeader { header, signature }) = futures::select! {
+        let (notary_encoder_seed, SignedSessionHeader { header, signature }) = futures::select_biased! {
```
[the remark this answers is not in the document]
themighty1: ok , i opened an issue to fix this https://github.com/tlsnotary/tlsn/issues/369

PR #362, opened 2023-10-13
Tidy up examples folder.

yuroitaki:
- Introduce folder structure for different examples
- Change simple_prover to only use simple_notary instead of notary server so that it's easier to understand
- Make sure both simple_notary and notary server use the same signing key
- Some modularisation to avoid repeated code


heeckhau (approved):
Nice improvements! Good work

tlsn/examples/discord/discord_dm.rs
```diff
@@ -251,9 +250,9 @@ async fn setup_notary_connection() -> (tokio_rustls::client::TlsStream<TcpStream
     .unwrap();
 
     let request = Request::builder()
-        .uri(format!("https://{NOTARY_DOMAIN}:{NOTARY_PORT}/session"))
+        .uri(format!("https://{NOTARY_HOST}:{NOTARY_PORT}/session"))
         .method("POST")
-        .header("Host", NOTARY_DOMAIN.clone())
+        .header("Host", NOTARY_HOST.clone())
```
themighty1, line 255: This Host header will be ignored by the server since the host is already provided in the .uri()

 https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html sec. 5.2 
"1. If Request-URI is an absoluteURI, the host is part of the Request-URI. Any Host header field value in the request MUST be ignored. "

we can just drop this header, i think

themighty1 (approved):
lgtm, thanks

th4s (approved):
nice :+1:

tlsn/examples/discord/discord_dm.rs
yuroitaki: @themighty1 ah but when I opened the link attached, it says it's superseded by this newer [rfc](https://www.rfc-editor.org/rfc/rfc9110#section-7.2), whose host section says

> A user agent MUST generate a Host header field in a request unless it sends that information as an ":authority" pseudo-header field. A user agent that sends Host SHOULD send it as the first field in the header section of a request.

Also found this from [Mozilla website](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)

> A Host header field must be sent in all HTTP/1.1 request messages. A [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) (Bad Request) status code may be sent to any HTTP/1.1 request message that lacks or contains more than one Host header field.

themighty1: thanks for checking, TIL

Issue #348, opened 2023-09-22
Repair bench for `prf` crate


th4s:
Currently the bench is broken with the following error
```
Benchmarking prf/prf: Warming up for 3.0000 sthread 'main' panicked at hmac-sha256/benches/prf.rs:29:6:
called `Result::unwrap()` on an `Err` value: InvalidState(Initialized)
```

This should be repaired.

0xvon:
Hi I'm James, talking with you on discord. Let me try to work on this issue this week!

0xvon:
@th4s 

I created a PR. Please review it when you have time.

https://github.com/tlsnotary/tlsn/pull/363

PR #363, opened 2023-10-14
Fix PRF benchmark

0xvon:
## Issue

https://github.com/tlsnotary/tlsn/issues/348

## Error

when call `cargo bench` at `./components/prf/hmac-sha256`, panicked with these messages:
```
Benchmarking prf/prf: Warming up for 3.0000 sthread 'main' panicked at hmac-sha256/benches/prf.rs:29:6:
called `Result::unwrap()` on an `Err` value: InvalidState(Initialized)
```

## Cause

The error caused since the state was still `Initialized`.

In the commit https://github.com/tlsnotary/tlsn/commit/329858b0fd25fed8928ebab622d5390253737de0 , we separated the building session keys as setup() function. So, we must call the method `setup()` before calling each `compute_xx()` methods.

## What I did on this PR

added this line

```diff
    let mut leader = MpcPrf::new(leader_vm.new_thread("bench_thread").await.unwrap());
    let mut follower = MpcPrf::new(follower_vm.new_thread("bench_thread").await.unwrap());

+    futures::try_join!(leader.setup(), follower.setup()).unwrap();

    let pms = [42u8; 32];
```

## How to review

1. call `cargo bench` at `./components/prf/hmac-sha256`.
2. make sure bench works.

### DIFF
--- a/components/prf/hmac-sha256/benches/prf.rs
+++ b/components/prf/hmac-sha256/benches/prf.rs
@@ -9,6 +9,8 @@ async fn bench_prf() {
     let mut leader = MpcPrf::new(leader_vm.new_thread("bench_thread").await.unwrap());
     let mut follower = MpcPrf::new(follower_vm.new_thread("bench_thread").await.unwrap());
 
+    futures::try_join!(leader.setup(), follower.setup()).unwrap();
+
     let pms = [42u8; 32];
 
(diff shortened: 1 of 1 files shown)


th4s:
Thank you very much :pray:

th4s (approved):

PR #366, opened 2023-10-17
Modify notary url parsing to make it generic for both ip and domain

yuroitaki:
Make it generic to parse notary url in the prover example — so that we can just easily use ip address or domain name as the notary host


heeckhau (approved):
👍

PR #375, opened 2023-10-25
`tlsn-formats`: fix error on duplicate headers

### LINKED ISSUE #374: `tlsn-formats` commitment builders fail in presence of duplicate headers
I think I've found an issue with current commitment builder (`HttpResponseCommitmentBuilder` in my case to be precise): `Set-Cookie` header can be set multiple times, but it causes `.headers()` call to end with `"attempted to create a duplicate commitment, overwriting: CommitmentId(16)"` error.

Here's a failing test, put it to `tlsn/tlsn-formats/src/http/mod.rs`:
```rust
#[test]
fn test_http_commit_duplicate_headers() {
    let tx: &[u8] = b"GET / HTTP/1.1\r\nHost: localhost\r\n\r\n";
    let rx: &[u8] = b"HTTP/1.1 200 OK\r\nSet-Cookie: lang=en; Path=/\r\n\
    Set-Cookie: fang=fen; Path=/\r\nContent-Length: 14\r\n\r\n{\"foo\": \"bar\"}";
    let mut transcript_commitment_builder = TranscriptCommitmentBuilder::new(
        fixtures::encoding_provider(tx, rx),
        tx.len(),
        rx.len(),
    );

    let requests = parse_requests(Bytes::copy_from_slice(tx)).unwrap();
    let responses = parse_responses(Bytes::copy_from_slice(rx)).unwrap();

    HttpCommitmentBuilder::new(&mut transcript_commitment_builder, &requests, &responses)
        .build()
        .unwrap();

    let commitments = transcript_commitment_builder.build().unwrap();

    // Path
    assert!(commitments
        .get_id_by_info(CommitmentKind::Blake3, (4..5).into(), Direction::Sent)
        .is_some());
    // Host
    assert!(commitments
        .get_id_by_info(CommitmentKind::Blake3, (22..31).into(), Direction::Sent)
        .is_some());

    // Set-Cookie 1
    assert!(commitments
        .get_id_by_info(CommitmentKind::Blake3, (29..44).into(), Direction::Received)
        .is_some());
    // Set-Cookie 2
    assert!(commitments
        .get_id_by_info(CommitmentKind::Blake3, (58..74).into(), Direction::Received)
        .is_some());
}
```

```
...
running 1 test
thread 'http::tests::test_http_commit_duplicate_headers' panicked at tlsn-formats/src/http/mod.rs:225:14:
called `Result::unwrap()` on an `Err` value: Commitment(Duplicate(CommitmentId(4)))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test http::tests::test_http_commit_duplicate_headers ... FAILED
...
```

rozag:
This resolves #374

### DIFF
--- a/tlsn/tlsn-formats/src/http/commitment.rs
+++ b/tlsn/tlsn-formats/src/http/commitment.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 
 use crate::http::{Body, BodyCommitmentBuilder, Request, Response};
-use spansy::Spanned;
+use spansy::{http::Header, Spanned};
 use tlsn_core::{
     commitment::{CommitmentId, TranscriptCommitmentBuilder, TranscriptCommitmentBuilderError},
     Direction,
@@ -156,7 +156,13 @@ impl<'a> HttpRequestCommitmentBuilder<'a> {
             .0
             .header(name)
             .ok_or(HttpCommitmentBuilderError::MissingHeader(name.to_string()))?;
+        self.header_internal(header)
+    }
 
+    fn header_internal(
+        &mut self,
+        header: &Header,
+    ) -> Result<CommitmentId, HttpCommitmentBuilderError> {
         let range = header.value.span().range();
         let id = self.builder.commit_sent(range.clone())?;
 
@@ -173,7 +179,7 @@ impl<'a> HttpRequestCommitmentBuilder<'a> {
 
         for header in &self.request.0.headers {
             let name = header.name.span().as_str().to_string();
-            let id = self.header(&name)?;
+            let id = self.header_internal(header)?;
 
             commitments.push((name, id));
         }
@@ -210,7 +216,7 @@ impl<'a> HttpRequestCommitmentBuilder<'a> {
 
             let range = header.value.span().range();
             if !range.is_subset(&self.committed) {
-                self.header(&name)?;
+                self.header_internal(header)?;
             }
         }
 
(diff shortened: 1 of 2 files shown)


th4s (commented):
Thanks a lot for the PR!

While this change is helpful, because it makes sure that the "commit everything" approach (which is currently default) works, I think we can improve it further.

The current approach does not provide a solution for the case where I want to manually commit to 2 `Set-Cookie` header fields by calling `HttpRequestCommitmentBuilder::header`.

The reason is that the spansy [Request](https://github.com/sinui0/spansy/blob/main/src/http/types.rs#L92-L95) and [Response](https://github.com/sinui0/spansy/blob/main/src/http/types.rs#L135-L138) types assume that a header is unique.

If we remove this assumption in spansy, then we would only need to adapt the API here to reflect this change. I feel that this would be a cleaner solution.

rozag:
Oops, that's right, I've totally missed that use-case.

rozag:
@th4s I've pushed updates with a dependency on `spansy` [PR](https://github.com/sinui0/spansy/pull/5). It's still a draft b/c we'd need to merge that one first, but I'd appreciate if you could take a look.

th4s (approved):
Thanks :+1:

themighty1 (changes requested):
Thanks for the PR!
We should make sure we don't commit to headers with empty values in `.headers()`

But also seems like calling `.header()` with an empty header should return an error. 
@sinui0, are you ok with returning an error there or do you have some better approach in mind ?

rozag:
@themighty1 I'm not sure I understand what changes are requested here. This PR focuses on duplicate headers, headers with empty values are covered in another one: https://github.com/tlsnotary/tlsn/pull/378
Maybe I should change something in that one?

themighty1 (approved):
Ack

PR #376, opened 2023-10-25
fix(formats): skip committing to empty header values

sinui0:
This PR fixes a bug which causes an error when empty header values are present in an HTTP record.

# Changes
- Skip committing to header values which are empty


th4s (commented):
Shouldn't we also adapt the `Http{Request, Response}CommitmentBuilder::header` methods?

PR #380, opened 2023-10-31
refactor: selective disclosure api

sinui0:
This PR redesigns our selective disclosure API, inverting control over the parsed data structures, passing them out to the user instead.

All previous data format structures have been ripped out in favor of re-exporting types from `spansy`, and then leveraging the `ToRangeSet` trait from `tlsn-utils` in the `tlsn-core` API. Users who want to implement their own format types can if they wish, their types only need to implement `ToRangeSet` to integrate with `tlsn-core`.

It introduces two new traits, `HttpCommit` and `JsonCommit` which a user can either implement themselves, or just use our default committer. Both traits provide blanket impls for all the methods, so a user can override something without having to reimplement everything.

This API is by no means final. It will continue to evolve as we introduce new commitment types. There is also room for new abstractions for generating _proofs_ similar to how we do commitments via the `*Commit` traits. For now I've kept that out of scope of this PR and users will have to use the `SubstringsProofBuilder` API directly (which isn't too bad because at least now they can use typed data instead of manually specifying ranges).

### DIFF
--- a/tlsn/tlsn-core/src/proof/substrings.rs
+++ b/tlsn/tlsn-core/src/proof/substrings.rs
@@ -23,7 +23,7 @@ pub enum SubstringsProofBuilderError {
     /// Invalid commitment id.
     #[error("invalid commitment id: {0:?}")]
     InvalidCommitmentId(CommitmentId),
-    /// Missing commitment
+    /// Missing commitment.
     #[error("missing commitment")]
     MissingCommitment,
     /// Invalid commitment type.


themighty1:
ack on the concept

rozag:
Nice! Looking forward to switching to this new API 👍

th4s (approved):
Yes I think this is a good approach :+1:

tlsn/tlsn-formats/src/json/commit.rs
```diff
@@ -0,0 +1,250 @@
...
+                )
+            })?;
+
+        self.commit_value(builder, &kv.value, direction)
+    }
+
+    /// Commits a JSON array.
+    ///
+    /// The default implementation commits to the entire array, then commits the array
+    /// excluding all values.
```
themighty1, line 135:
```suggestion
    /// excluding all values and separators.
```

themighty1 (changes requested):
looks good,

can we use the same formatting for method args everywhere in this PR as per https://github.com/tlsnotary/tlsn/blob/dev/CONTRIBUTING.md#function-arguments

/// * `arg1` - The first argument.
(The first letter capitalized and a period at the end)

tlsn/tlsn-formats/src/http/commit.rs
```diff
@@ -0,0 +1,387 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `builder` - the transcript commitment builder
+    /// * `direction` - the direction of the request (sent or received)
+    /// * `request` - the request to commit to
+    fn commit_request(
+        &mut self,
+        builder: &mut TranscriptCommitmentBuilder,
+        direction: Direction,
```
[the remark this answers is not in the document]
themighty1: ok, the server can be the prover, understood. Then we need to change the doc comment for `enum Direction`. Currently it says     
/// Sent from the prover to the server

tlsn/tlsn-formats/src/json/commit.rs
```diff
@@ -0,0 +1,250 @@
...
+                JsonCommitError::new_with_source("failed to commit array excluding values", e)
+            })?;
+
+        // TODO: Commit each value separately, but we need a strategy for handling
+        // separators.
+
+        Ok(())
+    }
+
+    /// Commits to a JSON string.
```
themighty1, line 164:
```suggestion
    /// Commits to an unquoted JSON string.
```

themighty1 (approved):
gw

tlsn/tlsn-formats/src/json/commit.rs
```diff
@@ -0,0 +1,250 @@
...
+            .map_err(|e| JsonCommitError::new_with_source("failed to commit array", e))?;
+
+        builder
+            .commit(&array.without_values(), direction)
+            .map_err(|e| {
+                JsonCommitError::new_with_source("failed to commit array excluding values", e)
+            })?;
+
+        // TODO: Commit each value separately, but we need a strategy for handling
+        // separators.
```
th4s, line 159: This is probably part of a follow-up PR later?

th4s (approved):
:rocket: gw. Very nice API.

Issue #367, opened 2023-10-18
Add feature flag to enable the usage of `web-time` instead of `std::time`


heeckhau:
Some methods of `std::time` panic when they are called when using the `wasm32-unknown-unknown` target.

If this new flag is set, [`web-time`](https://lib.rs/crates/web-time) should be used instead.

See https://github.com/tlsnotary/tlsn/compare/dev...mhchia:tlsn:dev-20231006-webtime-2

heeckhau:
https://lib.rs/crates/web-time:
> At the same time the library will simply re-export [std::time](https://doc.rust-lang.org/stable/std/time/) when not using the wasm32-unknown-unknown target and will not pull in any dependencies.

@sinui0 OK to just use `web-time` instead of `time`?  If the above is true, there is not need for a feature flag.

heeckhau:
Fixed in #383

PR #383, opened 2023-11-02
Replace `std::time::SystemTime::now` to support wasm32

mhchia:
## Overview
There is an issue with `std::time::SystemTime::now` not being supported in the `wasm32` target, as reported in https://github.com/tlsnotary/tlsn/issues/367.

## Changes Made
- Replaced necessary `std::time`s with `web_time`. Only left some `std::time` which are used only in tests and are not related to our changes.

## Discussion Points

Q1: Considering that `instant` is now utilized for compatibility, should we proactively update all other instances of `std::time::{Duration, SystemTime, Instant}` to their `instant` counterparts for consistency?

Q2: To make `tlsn-extension` with `tlsn`, we need to specify `instant = { version = "0.1", features = ["wasm-bindgen"] }` in `tlsn-extension`'s dependencies. I'm thinking we can avoid doing this in `tlsn-extension` by adding the following settings
```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
instant = { workspace = true, features = ["wasm-bindgen"] }
```
in both `tlsn/tlsn-core/Cargo.toml` and `tlsn/tlsn-prover/Cargo.toml`. Is this a suitable method, or are there better alternatives? Thanks!

### DIFF
--- a/tlsn/tlsn-prover/src/tls/mod.rs
+++ b/tlsn/tlsn-prover/src/tls/mod.rs
@@ -20,7 +20,6 @@ use rand::Rng;
 use std::{pin::Pin, sync::Arc};
 use tls_client_async::{bind_client, ClosedConnection, TlsConnection};
 use tls_mpc::{setup_components, MpcTlsLeader, TlsRole};
-use web_time::SystemTime;
 
 use mpz_garble::{config::Role as DEAPRole, protocol::deap::DEAPVm};
 use mpz_ot::{


mhchia:
@sinui0 Thanks for the answer! Sorry that I didn't notice the issue either. 😢 In this case I agree `web-time` should be better. I will replace all `std::time` with `web_time`. Since `web_time` already checks wasm32 for us, we won't need to do it in tlsn.

th4s (approved):
Nice :+1:

themighty1 (approved):
lgtm

Issue #330, opened 2023-09-13
Support deferred decryption


sinui0:
Support deferring decryption of server response data until *after the TLS connection is closed*. Once the connection is closed, the Verifier can reveal their share of the TLS session keys to the Prover. The Prover can then decrypt the payload without using full MPC, but rather just the ZK phase of DEAP. This is a significant optimization to total protocol time as it eliminates the majority of the *upload bandwidth* constraint.

ameya-deshmukh:
@sinui0 can I take this up?

ameya-deshmukh:
Sure, will do!

PR #384, opened 2023-11-07
feat: stream cipher zk

sinui0:
This PR simplifies our stream cipher implementation quite a bit and adds ZK functionality to its public API. This is the first step needed to support #330 and it also provides the changes needed to bump `mpz` version to support #379 

# Changes
- Modified AES-CTR circuit to only compute the encrypted counter block, without directly applying the input text.
- Removed padding of the input text, this is no longer needed as we apply the keystream to the input text using a dynamically generated XOR circuit.
- Added the `prove_plaintext` and `verify_plaintext` methods
- General code clean up, refactoring out complexity


th4s (approved):
nice :+1:

themighty1 (approved):
gw, ack

PR #385, opened 2023-11-14
Make CORS permissive

mhchia:
Disable CORS as the notary server is meant to be consumed by all origins.

cc @yuroitaki  @0xtsukino


yuroitaki (changes requested):
Looks good! Just need to run cargo fmt and merge in the latest dev

PR #393, opened 2023-12-12
Optimise test to use release profile to speed up notary server integration test

yuroitaki:
Optimise test to use release profile to speed up notary server integration test as it was taking close to 2 mins


heeckhau (approved):
👍

PR #394, opened 2023-12-14
Add ci build for WASM target for specific packages

heeckhau:


th4s (approved):

tlsn/tlsn-core/Cargo.toml
```diff
@@ -45,3 +45,6 @@ bincode.workspace = true
 [[test]]
 name = "api"
 required-features = ["fixtures"]
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
mhchia, line 49: Should `ring = { version = "0.17", features = ["wasm32_unknown_unknown_js"] }` also be required here? Since `tlsn-tls-core` is a dependency and it depends on `ring`

heeckhau: @mhchia I am not sure. I have not fully analysed the dependencies.
It builds without specifying it here, so that is why I didn't add it here.

th4s: I think `tls-core` already pulls in `ring 0.17` because it has `ring.workspace = true`. So no need to make it a dependency of `tlsn-core`.

mhchia: @heeckhau I see 🙏
@th4s Yeah but in order for `tlsn-core` (and ring 0.17) to build in wasm32, should we specify the feature flag `wasm32_unknown_unknown_js` for `ring 0.17`, or do I misunderstand something?

th4s: Ahh, haven't thought about this. Following this [additive-features] https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification it would probably work in most cases, since it is never pulled in alone.

But it would probably be better to add the feature to `tls-core` then?

mhchia: > Following this [additive-features](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification) it would probably work in most cases, since it is never pulled in alone.

I see. Thanks for sharing! If `tlsn-core` is not used alone then whether adding or not both works for me 🙏

PR #403, opened 2024-01-09
Authdecode prototype

themighty1:
This is a draft PR integrating the Authdecode protocol into the notarization flow.


maceip:
SO STOKED

PR #405, opened 2024-01-10
MPC-TLS actor refactor

sinui0:
This PR refactors our MPC-TLS implementation to use `ludi` and generally have better state management, errors and logging.

# Changes
- Modified `Backend` trait to have all `async` methods with only owned arguments.
- Updated `tls-client` for changes to backend trait.
- Refactored `MpcTlsLeader` and `MpcTlsFollower` to use `ludi`
- `MpcTlsLeader` is no longer passed directly into the tls client, instead a controller is.
- Refactored the `MpcTlsError` to be more opaque, concealing implementation details that shouldn't be in the API. At the same time, it provides better debugging information.
- Introduced cleaner internal states to both the leader and follower, and added state guards in numerous places to catch unexpected events.
- Introduced more explicit protocol messages between the leader and follower.
- Encapsulated encryption functionality into `record_layer`
- Decoupled message commitment from message decryption, to support deferred decryption in a coming PR.

### DIFF
--- a/components/tls/tls-mpc/src/follower.rs
+++ b/components/tls/tls-mpc/src/follower.rs
@@ -81,8 +81,8 @@ impl ludi::Actor for MpcTlsFollower {
 
         if !committed.is_empty() {
             return Err(MpcTlsError::new(
-                Kind::PeerMisbehaved,
-                "leader attempted to finalize without proving all messages",
+                Kind::Other,
+                "not all committed TLS messages were verified",
             ));
         }
 


th4s (approved):
lgtm :+1:

themighty1:
thanks, will review soon

components/tls/tls-mpc/src/follower.rs
```diff
@@ -148,301 +269,370 @@ impl MpcTlsFollower {
...
+    pub(super) struct Sf {
+        pub(super) handshake_commitment: Option<Hash>,
+        pub(super) server_key: PublicKey,
+    }
+
+    #[derive(Debug)]
+    pub(super) struct Active {
+        pub(super) handshake_commitment: Option<Hash>,
+        pub(super) server_key: PublicKey,
+        pub(super) committed: VecDeque<OpaqueMessage>,
```
themighty1, line 627:
```suggestion
         /// TLS messages purportedly received by the leader from the server.
         ///
         /// The follower must verify the authenticity of these messages with AEAD verification
         /// (i.e. by verifying the authentication tag).
        pub(super) committed: VecDeque<OpaqueMessage>,
```

components/tls/tls-mpc/src/msg.rs
```diff
@@ -1,54 +1,123 @@
...
-    HandshakeCommitment(Hash),
+    ComputeClientKey(ComputeClientKey),
+    ComputeKeyExchange(ComputeKeyExchange),
+    ClientFinishedVd(ClientFinishedVd),
+    EncryptClientFinished(EncryptClientFinished),
+    EncryptAlert(EncryptAlert),
+    ServerFinishedVd(ServerFinishedVd),
+    DecryptServerFinished(DecryptServerFinished),
+    DecryptAlert(DecryptAlert),
+    CommitMessage(CommitMessage),
```
themighty1, line 36:
```suggestion
     /// A leader commitment to a TLS message received from the server.
    CommitMessage(CommitMessage),
```

themighty1 (approved):
ACK, left some non-blocking comments

PR #408, opened 2024-01-12
add RustCrypto TLS prf to tls-core

sinui0:
This PR adds the TLS1.2 PRF to `tls-core`, implemented using RustCrypto primitives and based on the ring implementation from rustls.


th4s (commented):
:+1:
Should we introduce some type checks considering the length of the slices, by using array references (static check) or `assert_eq` (runtime check). I feel this would prevent some possible error sources.

PR #410, opened 2024-01-16
style: change the name to NotarySigningKeyProperties

themighty1:


yuroitaki (commented):
Looks good! Just need to update those in the integration_test.rs as well

themighty1:
> Looks good! Just need to update those in the integration_test.rs as well

done

PR #415, opened 2024-01-16
refactor: add AEAD prove/verify + tag methods

sinui0:
This PR refactors the MPC AEAD a bit to provide methods for proving/verifying messages with or without a tag. It also encapsulates the tag verification a bit better.

I also adjusted the logging to not include message data while I was in there.

### DIFF
--- a/components/aead/src/aes_gcm/mod.rs
+++ b/components/aead/src/aes_gcm/mod.rs
@@ -177,7 +177,7 @@ impl MpcAesGcm {
             .compute_tag(explicit_nonce, payload.clone(), aad)
             .await?;
 
-        // Reject if tag is incorrect
+        // Reject if tag is incorrect.
         if tag != purported_tag {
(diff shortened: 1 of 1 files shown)


components/aead/src/aes_gcm/mod.rs
```diff
@@ -163,6 +163,27 @@ impl MpcAesGcm {
...
+        payload: &mut Vec<u8>,
+        aad: Vec<u8>,
+    ) -> Result<(), AeadError> {
+        let purported_tag = payload.split_off(payload.len() - AES_GCM_TAG_LEN);
+
+        let tag = self
+            .compute_tag(explicit_nonce, payload.clone(), aad)
+            .await?;
+
+        // Reject if tag is incorrect
```
themighty1, line 180:
```suggestion
        // Reject if tag is incorrect.
```

themighty1 (approved):

PR #416, opened 2024-01-21
feat: deferred decryption

### LINKED ISSUE #330: Support deferred decryption

sinui0:
This PR implements so-called "deferred decryption", and advanced feature which significantly reduces upload bandwidth utilization by the Prover. This should provide much faster proving times.

There is a remaining issue with this feature related to detecting when all the application data has been received while we are deferring decryption. If the Server does not close their end of the connection it will cause a deadlock. The remedy is to have the Prover send a close_notify after they have sent their final request. This becomes a problem when trying to use `hyper` as it takes ownership of the connection which prevents closing it manually. I've punted implementing good workarounds for this out of scope of this PR.

Closes #330 

# Changes
- Added additional control flow functions to the `Backend` trait
- Major refactors to `tls-client-async` to make this feature possible. I also did some overall refactoring to simplify it which should cut out some bugs and make it more readable.
- Added the functionality to `tls_mpc`.

### DIFF
--- a/components/tls/tls-mpc/src/leader.rs
+++ b/components/tls/tls-mpc/src/leader.rs
@@ -342,11 +342,6 @@ impl MpcTlsLeader {
 
 #[ludi::implement(msg(name = "{name}"), ctrl(err))]
 impl MpcTlsLeader {
-    /// Returns whether messages are being decrypted.
-    pub async fn is_decrypting(&self) -> Result<bool, MpcTlsError> {
-        Ok(self.is_decrypting)
-    }
-
     /// Closes the connection.
     #[cfg_attr(
         feature = "tracing",
--- a/components/tls/tls-mpc/src/msg.rs
+++ b/components/tls/tls-mpc/src/msg.rs
@@ -17,7 +17,7 @@ use crate::{
         BackendMsgSetCipherSuite, BackendMsgSetDecrypt, BackendMsgSetEncrypt,
         BackendMsgSetHsHashClientKeyExchange, BackendMsgSetHsHashServerHello,
         BackendMsgSetProtocolVersion, BackendMsgSetServerCertDetails, BackendMsgSetServerKeyShare,
-        BackendMsgSetServerKxDetails, BackendMsgSetServerRandom, DeferDecryption, IsDecrypting,
+        BackendMsgSetServerKxDetails, BackendMsgSetServerRandom, DeferDecryption,
     },
     MpcTlsError,
 };
@@ -96,7 +96,6 @@ pub enum MpcTlsLeaderMsg {
     BackendMsgGetNotify(BackendMsgGetNotify),
     BackendMsgBufferLen(BackendMsgBufferLen),
     BackendMsgServerClosed(BackendMsgServerClosed),
-    IsDecrypting(IsDecrypting),
     DeferDecryption(DeferDecryption),
     CloseConnection(CloseConnection),
     Finalize(Commit),


th4s (approved):
:fire: lgtm

components/tls/tls-mpc/src/leader.rs
```diff
@@ -327,33 +356,38 @@ impl MpcTlsLeader {
...
+
+        self.is_decrypting = false;
+        self.notifier.clear();
+
+        Ok(())
+    }
+
+    /// Commits the leader to the current transcript.
+    ///
+    /// This reveals the encryption keys to the leader and disables sending or receiving
```
themighty1, line 382:
```suggestion
    /// This reveals the AEAD key to the leader and disables sending or receiving
```

components/tls/tls-mpc/src/leader.rs
```diff
@@ -104,7 +110,10 @@ impl MpcTlsLeader {
             prf,
             encrypter,
             decrypter,
-            committed: VecDeque::new(),
+            notifier: BackendNotifier::new(),
+            is_decrypting: true,
```
themighty1, line 114: isnt `is_decrypting` just a reflection of the `notifier`'s state? Here they are set in the opposite states. Doesnt seem like they should be.

components/tls/tls-mpc/src/leader.rs
```diff
@@ -639,13 +673,52 @@ impl Backend for MpcTlsLeader {
...
+                .send(MpcTlsMessage::CommitMessage(CommitMessage {
+                    msg: msg.payload.0.clone(),
+                }))
+                .await
+                .map_err(|e| BackendError::InternalError(e.to_string()))?;
+        }
+
+        self.buffer.push_back(msg);
+
+        if self.is_decrypting {
```
themighty1, line 693: Note to self: same q here as above about what the purpose of `is_decrypting` is.

themighty1 (changes requested):
need to resolve some qs around the code logic

components/tls/tls-client-async/src/lib.rs
```diff
@@ -74,165 +72,163 @@ impl Future for ConnectionFuture {
...
+                #[cfg(feature = "tracing")]
+                trace!("forwarded {} plaintext bytes to conn", read);
+            }
+
+            if !client.is_handshaking() && !handshake_done {
+                #[cfg(feature = "tracing")]
+                debug!("handshake complete");
+                handshake_done = true;
+                // Start reading application data that needs to be transmitted from the `TlsConnection`.
+                tx_recv_fut = tx_receiver.next().fuse();
```
themighty1, line 142: Seems like with the current logic it is possible to silently lose some tx `data`.

1. Consider if some tx `data` is available at this point, then `tx_recv_fut` should resolve to that `data`.

2. Consider if in the `select_biased` loop below only the `rx_tls_fut` branch will be executed because there is some incoming data from the server (`select_biased!` executes only one branch iiuc). This means that the `tx_recv_fut` branch will **not** be executed and the `data` will **not** be sent.

3. In the next loop iteration `tx_recv_fut` on this line will be set to a future which resolves to something else, e.g. `None`. This will result in `data` not getting sent.

themighty1 (changes requested):
a q needs to be resolved

components/tls/tls-mpc/src/leader.rs
```diff
@@ -57,8 +59,14 @@ pub struct MpcTlsLeader {
     encrypter: Encrypter,
     decrypter: Decrypter,
 
+    /// Allows us to notify whether or not we are ready to decrypt TLS messages.
```
themighty1, line 62: sry, this new comment is less ambiguous, imo
```suggestion
    /// When set, notifies the backend that there are TLS messages which need to be decrypted.
```

components/tls/tls-mpc/src/leader.rs
```diff
@@ -57,8 +59,14 @@ pub struct MpcTlsLeader {
...
+    notifier: BackendNotifier,
+
+    is_decrypting: bool,
```
themighty1, line 65:
```suggestion
    /// Whether the backend is ready to decrypt messages.
    is_decrypting: bool,
```

components/tls/tls-mpc/src/follower.rs
```diff
@@ -216,6 +215,24 @@ impl MpcTlsFollower {
         Ok(())
     }
 
+    fn accepting_messages(&self) -> Result<(), MpcTlsError> {
+        if self.close_notify {
+            return Err(MpcTlsError::new(
+                Kind::PeerMisbehaved,
+                "attempted to commit a message after receiving CloseNotify",
```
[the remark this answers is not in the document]
themighty1: The problem with "commit" is that the Leader themselves don't know if they are "committing" to an authentic ciphertext until the MAC is verified.
Usually someone commits to a value that they know is authentic.

From the TLS-MPC client's point of view, it can be simply called "Receiving the TLS message". The fact that the leader receives it first and forwards it to the follower does not add any commitment properties to it, imo.
Both parties still need to authenticate the received message.

themighty1 (approved):
gw, some qs still need to be resolved

PR #418, opened 2024-01-24
perf: adjust mux config

sinui0:
This PR adjusts the multiplexer config to increase the receive window. This will reduce the effects of throttling as latency increases and should give us a theoretical throughput of ~300Mbps @ 60ms per stream instead of ~33Mbps.

The trade-off here is that it increases the memory allocated to buffers. Due to this, I capped the max streams to 40 to prevent DoS (our protocol uses a static number of streams, 33 atm). The existing default configuration provides a max of 8192 streams, which works out to ~2Gb maximum allocation if they were all opened. With the new settings the maximum allocation is 640Mb.

Newer versions of the `yamux` crate scale the window dynamically based on latency + number of opened streams. But their API has breaking changes that we need to fix if we want to bump later.

### DIFF
--- a/tlsn/tlsn-prover/src/tls/mod.rs
+++ b/tlsn/tlsn-prover/src/tls/mod.rs
@@ -75,6 +75,7 @@ impl Prover<state::Initialized> {
         socket: S,
     ) -> Result<Prover<state::Setup>, ProverError> {
         let mut mux_config = yamux::Config::default();
+        // See PR #418
         mux_config.set_max_num_streams(40);
         mux_config.set_max_buffer_size(16 * 1024 * 1024);
         mux_config.set_receive_window(16 * 1024 * 1024);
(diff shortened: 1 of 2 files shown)


th4s (approved):
:+1:

yuroitaki (approved):
LGTM!

tlsn/tlsn-prover/src/tls/mod.rs
```diff
@@ -74,7 +74,11 @@ impl Prover<state::Initialized> {
         self,
         socket: S,
     ) -> Result<Prover<state::Setup>, ProverError> {
-        let mut mux = UidYamux::new(yamux::Config::default(), socket, yamux::Mode::Client);
+        let mut mux_config = yamux::Config::default();
+        mux_config.set_max_num_streams(40);
```
heeckhau, line 78: I feel like we should add a comment here explaining the "magic numbers".
A link to this PR might be OK

themighty1 (approved):

PR #419, opened 2024-01-25
Deployment of notary-server using GH Actions & AWS CodeDeploy

ntampakas:
This PR introduces automatic deployment of notary-server binary to a dedicated AWS EC2 instance using GH Actions Workflow and triggering AWS CodeDeploy with temporary security credentials in IAM.
- The **cargo build** step is executed on a GH runner to avoid computational needs on our server and potential disruptions. This approach also helps to decrease app downtime.
- The new binary is uploaded to an S3 bucket, and when CodeDeploy is triggered it:
  - checkouts the selected branch
  - retrieves the binary from the S3 bucket
  - copies it to the appropriate directory
  - restarts the process.

Additionally, a log file has been introduced under the **tlsn** directory and the workflow supports nightly and stable versions deployment.


.github/scripts/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ex
+
+environment=$1
+
+cd notary-server
+cargo build --release
+aws s3 cp target/release/notary-server s3://tlsn-deploy/$environment/
```
heeckhau, line 8: How is this bucket  managed?

We could also use https://github.com/actions/upload-artifact
which offers the ` retention-days:` option.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment"
+        required: true
+        default: "enable"
+        type: choice
+        options:
+        - nightly
+        - stable
```
heeckhau, line 14: This is unused for now?

scripts/after_install.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+export PATH=$PATH:/home/ubuntu/.cargo/bin
+
+APP_NAME=$(echo $APPLICATION_NAME | awk -F- '{ print $2 }')
+BRANCH=$(curl http://169.254.169.254/latest/meta-data/tags/instance/$APP_NAME)
```
heeckhau, line 6: What server is this? Can we use a name here?

heeckhau (commented):
Looking good!
I have added some comments and questions

.github/scripts/build.sh
ntampakas: This resource was created and it' s managed by Terraform.
There is no lifecycle policy at the moment, but it wouldn' t be a problem to create one.

.github/workflows/deploy.yml
ntampakas: This option it' s being used and provides you with the selection of the environment you want to deploy:
nightly and stable

scripts/after_install.sh
ntampakas: 169.254.169.254 is used as the endpoint for the instance metadata service. This is an internal IPv4 address, and every EC2 instance has this IP address.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
+name: Deploy
+on:
```
yuroitaki, line 2: Can we have the following triggers set?
For nightly, it should be deployed upon push on dev branch
For stable, it should be deployed upon push on tag "[v]?[0-9]+.[0-9]+.[0-9]+*"

yuroitaki: Oh both should only run after ci.yml succeeds,. refer to https://github.com/tlsnotary/tlsn/blob/dev/.github/workflows/cd.yml#L19-L27

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
+name: Deploy
```
yuroitaki, line 1: Can we rename this and the filename to something clearer, i.e. Deploy to AWS, cd-aws.yml

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+          persist-credentials: false
+          
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::490752553772:role/tlsn-deploy-slc
+          role-duration-seconds: 1800
+          aws-region: eu-central-1
+
+      - name: Cargo build
```
yuroitaki, line 42: Prior to this step, do we need to install the same Rust version (i.e. stable) as that in the ci.yml step here: https://github.com/tlsnotary/tlsn/blob/dev/.github/workflows/ci.yml#L60-L72 — to ensure the same Rust version is being used for building?

Also we used Swatinem/rust-cache@v2.5.0 in ci, will that help here to speeding up deployment?

yuroitaki: I also realised in ci.yml, we have used the same build command `cargo build --release` for notary server, if there is an easy way to utilise the built artefact there, then it ll be great — it's fine if it's too troublesome

.github/scripts/deploy.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
yuroitaki, line 1: Can we rename this filename to something clearer, i.e. deploy-aws.sh

.github/scripts/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
yuroitaki, line 1: Can we rename this filename to something clearer, i.e. build-aws.sh

scripts/after_install.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
yuroitaki, line 1: Can we rename this folder name from `scripts` to `appspec-scripts` to make it clearer?

.github/scripts/build.sh
yuroitaki: Can we add a comment here very briefly explaining why we are doing this S3 binary copy step?

scripts/after_install.sh
```diff
@@ -0,0 +1,22 @@
...
+APP_NAME=$(echo $APPLICATION_NAME | awk -F- '{ print $2 }')
+BRANCH=$(curl http://169.254.169.254/latest/meta-data/tags/instance/$APP_NAME)
+ 
+# Prepare directory
+sudo rm -rf ~/$APP_NAME/tlsn
+sudo mv ~/tlsn/ ~/$APP_NAME
+sudo mkdir -p ~/$APP_NAME/tlsn/notary-server/target/release
+sudo chown -R ubuntu.ubuntu ~/$APP_NAME
+ 
+git clone -b $BRANCH --no-checkout https://github.com/tlsnotary/tlsn.git /tmp/tlsn_remove
```
yuroitaki, line 14: Can we add a comment here very briefly explaining why this step is needed?

P/S: there is a slim chance that this copied .git folder of the dev branch has included a newer commit that is not part of the built binary

appspec.yml
```diff
@@ -0,0 +1,30 @@
+version: 0.0
+os: linux
+files:
+  - source: /
```
yuroitaki, line 4: Does this mean that CodeDeploy ll always copy the source files to this destination regardless of stable or nightly? If both the cd pipelines of stable and nightly run together, will there be a race condition where potentially wrong source files are copied to wrong directory in after_install.sh line 10? (e.g. stable files got copied into nightly folder)?

yuroitaki (changes requested):
Looks great! Got a few comments, though most are just cosmetic

appspec.yml
```diff
@@ -0,0 +1,30 @@
+version: 0.0
```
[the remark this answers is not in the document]
ntampakas: Done! (https://github.com/ntampakas/tlsn/commit/c0c52bf55916e24ffe12f379df5b5a120a7babd2)

scripts/after_install.sh
ntampakas: Done! https://github.com/ntampakas/tlsn/commit/c0c52bf55916e24ffe12f379df5b5a120a7babd2

.github/workflows/deploy.yml
ntampakas: Done! https://github.com/ntampakas/tlsn/commit/e9776aa45d9adc35a597b4f24000bb08bb51ca79

.github/workflows/deploy.yml
ntampakas: Done! https://github.com/ntampakas/tlsn/commit/e9776aa45d9adc35a597b4f24000bb08bb51ca79

.github/scripts/build.sh
ntampakas: Done! https://github.com/ntampakas/tlsn/commit/6290dc85af85eb4857edd9ff457271a4a6d63e12

.github/scripts/deploy.sh
ntampakas: Done! https://github.com/ntampakas/tlsn/commit/6290dc85af85eb4857edd9ff457271a4a6d63e12

.github/workflows/deploy.yml
ntampakas: Done! (https://github.com/ntampakas/tlsn/commit/33641e1c614c43e74fd4854e4a21d32eda81ddf1)

Well, we could move the S3 copy step from the **cd-server.yml** workflow to the **ci.yml** one. This requires some modifications to the ci.yml workflow, such as using temporary security credentials and injecting the environment as well. If you are okay with it, I can start working on it.

ntampakas: > Prior to this step, do we need to install the same Rust version (i.e. stable) as that in the ci.yml step here: https://github.com/tlsnotary/tlsn/blob/dev/.github/workflows/ci.yml#L60-L72 — to ensure the same Rust version is being used for building?
> 
> Also we used Swatinem/rust-cache@v2.5.0 in ci, will that help here to speed up deployment?

Done! (https://github.com/ntampakas/tlsn/commit/33641e1c614c43e74fd4854e4a21d32eda81ddf1)

scripts/after_install.sh
ntampakas: If you are ok with this: https://github.com/tlsnotary/tlsn/pull/419#discussion_r1469599960
i can also migrate this part to ci.yml workflow.

ntampakas: Either what you suggested or if you agree with this https://github.com/tlsnotary/tlsn/pull/419#discussion_r1469599960
i can migrate the .github part to ci.yml workflow.

Either what you suggested or if you agree with this, i can migrate the .github part to ci.yml workflow.

ntampakas: Either what you suggested or if you agree with this https://github.com/tlsnotary/tlsn/pull/419#discussion_r1469599960
i can migrate the .github part to ci.yml workflow.

.github/scripts/build.sh
ntampakas: We use an S3 bucket because it' s the most straightforward and effective method for our case. Another approach could be uploading artifacts to GH using that specific action mentioned above, which might be convenient if you want to directly access the binary from this particular workflow. However, if you retrieve it from a 3rd party instance, it becomes more complicated. In our case, the workflows are executed on GH runners and have the necessary permissions to trigger deployment against our hosted AWS instance. To achieve this using GH artifacts, we would need to use the GH API and a token (running on the EC2 instance), which can be somehow complicated.
We can give it a try if you think it makes sense (the last time we attempted this was a year ago).

appspec.yml
ntampakas: I used this condition https://github.com/tlsnotary/tlsn/pull/419/files#diff-27ec5655b90f6efb2697e4cac232cc8ac76273ab346b84891a0f28c81931e171R19-R21
that ensures that only one workflow runs per concurrency group.
If a new run is triggered while another is still in-progress, the in-progress run is canceled to make way for the latest one.
If **cancel-in-progress: false**, the queued workflow will be pending and start right after the in-progress ends.

.github/workflows/cd-server.yml
```diff
@@ -0,0 +1,75 @@
...
+          role-duration-seconds: 1800
+          aws-region: eu-central-1
+
+      - name: Install stable rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
+          components: clippy
+
+      - name: "Clippy"
```
yuroitaki, line 61: we probably don't need this clippy check

appspec.yml
yuroitaki: ah nice! but that seems like it's making sure no jobs triggered by the same branch/tag run together — does it also ensure if a job is running for stable (i.e. upon push to tag), a new job for nightly (i.e. upon push to dev) will wait for the stable job to finish before running?

.github/workflows/deploy.yml
yuroitaki: Hmm I think it's better to keep them separate to avoid complicating the ci workflow file, so good with building it again in GH runner :)

scripts/after_install.sh
yuroitaki: let's go with my original suggestion if possible, so that the ci.yml workflow is kept simple and not too complicated :)

appspec-scripts/.start_app.sh.swp
yuroitaki, line 1: need this binary file?

.github/workflows/cd-server.yml
```diff
@@ -0,0 +1,75 @@
...
+    branches:
+      - dev
+    tags:
+      - "[v]?[0-9]+.[0-9]+.[0-9]+*"
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment"
+        required: true
+        default: "enable"
```
yuroitaki, line 13: Sorry can I know how does this trigger setting ensure that push to dev will set DATA_ENV as nightly; while push to tag will set DATA_ENV as stable? 🤔

.github/scripts/build.sh
yuroitaki: thanks for the detailed explanation! aha I actually meant to put a small link to this PR on the script

ntampakas: https://github.com/ntampakas/tlsn/commit/48a9f8e1b421b92d5145a4904b808d10a237617a

scripts/after_install.sh
ntampakas: Done! (https://github.com/ntampakas/tlsn/commit/022b491c1e8f92429dc430512167634f625489b2)

.github/workflows/cd-server.yml
ntampakas: Done! (https://github.com/ntampakas/tlsn/commit/022b491c1e8f92429dc430512167634f625489b2)

appspec-scripts/.start_app.sh.swp
ntampakas: File removed! (https://github.com/ntampakas/tlsn/commit/2cc9a05685dfebfc16dbc590557f375f2dea4ca3)

.github/workflows/cd-server.yml
ntampakas: Indeed, after introducing push on multiple branches, an environment manipulation step should be introduced in the workflow. I' ll push this change ASAP.

appspec.yml
ntampakas: Hm... you 're right. Just modified the concurrency group to the workflow level: https://github.com/ntampakas/tlsn/commit/445c1afa11482d39bc7c20a181fd22d3da2971d7

.github/workflows/cd-server.yml
ntampakas: https://github.com/ntampakas/tlsn/commit/5d9471c2645597f1b03fbf23630b97e61b69f0d9

.github/workflows/cd-server.yml
```diff
@@ -0,0 +1,87 @@
...
```
yuroitaki, line 13: @ntampakas may i know why this default value is 'enable'?

ntampakas: My bad @yuroitaki 
This is just a typo. Thanks for noticing it!
Fixed: https://github.com/ntampakas/tlsn/commit/56c4aa5074f9155fd9550bbde611348053c6a64b

yuroitaki (approved):
Thanks so much again for your help @ntampakas !

Issue #412, opened 2024-01-16
Document how to filter notary server logs


sinui0:
We should document how to filter notary server logs to only `tlsn` crates.

The notary server uses [`EnvFilter`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html) and can be configured using `RUST_LOG`. We should document how to use it.

heeckhau:
@yuroitaki can I assign this to you?

PR #422, opened 2024-02-06
Use logging filter, remove otel.

yuroitaki:
This addresses https://github.com/tlsnotary/tlsn/issues/412

- ~Switched from using config file to indicate log verbosity level to using RUST_LOG env var which is much more flexible~
- Removed opentelemetry setup code which is flooding the logs with `SpanData` struct because it was temporarily setup to export to stdout — we can remove it for now since we're not using any opentelemetry tool yet
- Fixed the config file logging setting to use logging filter directive instead of just verbosity level

### DIFF
--- a/components/tls/tls-client/src/backend/standard.rs
+++ b/components/tls/tls-client/src/backend/standard.rs
@@ -9,7 +9,7 @@ use p256::{ecdh::EphemeralSecret, EncodedPoint, PublicKey as ECDHPublicKey};
 use rand::{rngs::OsRng, thread_rng, Rng};
 
 use digest::Digest;
-use std::{any::Any, convert::TryInto, collections::VecDeque};
+use std::{any::Any, collections::VecDeque, convert::TryInto};
 use tls_core::{
     cert::ServerCertDetails,
     ke::ServerKxDetails,
--- a/components/tls/tls-client/src/backend/standard.rs
+++ b/components/tls/tls-client/src/backend/standard.rs
@@ -9,7 +9,7 @@ use p256::{ecdh::EphemeralSecret, EncodedPoint, PublicKey as ECDHPublicKey};
 use rand::{rngs::OsRng, thread_rng, Rng};
 
 use digest::Digest;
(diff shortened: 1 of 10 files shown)


notary-server/README.md
```diff
@@ -82,6 +82,18 @@ To perform notarization using the session id (unique id returned upon calling th
 ##### Query Parameter Type
 String
 
+---
+## Debugging
+The default logging of this server is set to `DEBUG` verbosity level for all crates (both internal and external) using tracing crate's [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html).
```
heeckhau, line 87: Maybe we should call this section `Logging`?

heeckhau (changes requested):

components/tls/tls-client/src/backend/standard.rs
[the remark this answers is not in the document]
yuroitaki: yep ran cargo fmt and affected them, but have excluded them

yuroitaki:
By the way @sinui0 @heeckhau please don't merge this first even after all approvals have been obtained when I am away — since there's a format change of the config file, I'd like to change that in our AWS server first before CD kicks in which ll fail otherwise (though that's actually okay since we can manually start the server after, just my preference to correct things before CD fails)

notary-server/README.md
```diff
@@ -82,6 +82,16 @@ To perform notarization using the session id (unique id returned upon calling th
 ##### Query Parameter Type
 String
 
+---
+## Logging
+The default logging strategy of this server is set to `DEBUG` verbosity level for the crates that are useful for most debugging scenarios, i.e. using the following filtering logic,
```
heeckhau, line 87:
```suggestion
The default logging strategy of this server is set to `DEBUG` verbosity level for the crates that are useful for most debugging scenarios, i.e. using the following filtering logic:
```

heeckhau (approved):

PR #428, opened 2024-02-08
fix(tls-core): use non_exhaustive instead of private zst

sinui0:
This PR fixes CI, satisfying clippy's complaint about using a private zst instead of non_exhaustive.


yuroitaki (approved):
LGTM

PR #433, opened 2024-02-09
feat(tlsn-formats): default commit to entire http request/response

sinui0:
This PR updates the default HTTP strategy to also commit to the entire request/response in one shot, for cases where nothing needs to be redacted.


tlsn/tlsn-formats/src/http/commit.rs
```diff
@@ -270,6 +278,14 @@ pub trait HttpCommit {
         direction: Direction,
         response: &Response,
     ) -> Result<(), HttpCommitError> {
+        builder.commit(response, direction).map_err(|e| {
```
themighty1, line 281: What if the HTTP response looks like:
`HTTP-Version SP Status-Code SP Reason-Phrase CRLF
CRLF`
then this line's commitment will be the same as `.without_data()` below and will cause a duplicate commitment error.

themighty1 (changes requested):

th4s (approved):

PR #435, opened 2024-02-11
feat: separate transcript size limits

### LINKED ISSUE #396: Accurately calculate MPC preprocessing
In the offline phase, we setup MPC protocols based on the `max_transcript_size` configuration. Right now the calculation for the number of OTs we need is naive (and wrong).

Investigate an accurate means for calculating how much preprocessing is required and update the corresponding configuration builders.

sinui0:
This PR separates the size limits for sent and received data to allow for better control over the setup costs of the protocol. Additionally, I've improved the logic for computing the initial OT extension size based on these limits. The estimates are not perfect, but are close enough while providing some extra padding.

This should reduce setup costs a bit, as we no longer provision a bunch of unneeded OTs from the Prover to the Verifier.

Closes #396

### DIFF
--- a/notary-server/tests/integration_test.rs
+++ b/notary-server/tests/integration_test.rs
@@ -34,6 +34,8 @@ use notary_server::{
 
 const NOTARY_CA_CERT_PATH: &str = "./fixture/tls/rootCA.crt";
 const NOTARY_CA_CERT_BYTES: &[u8] = include_bytes!("../fixture/tls/rootCA.crt");
+const MAX_SENT: usize = 1 << 13;
+const MAX_RECV: usize = 1 << 13;
 
 fn get_server_config(port: u16, tls_enabled: bool) -> NotaryServerProperties {
     NotaryServerProperties {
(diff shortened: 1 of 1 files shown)


tlsn/tlsn-common/src/config.rs
```diff
@@ -0,0 +1,44 @@
+//! TLSNotary protocol config and config utilities.
+
+use crate::Role;
+
+/// Default for the maximum number of bytes that can be sent (4Kb).
+pub const DEFAULT_MAX_SENT_LIMIT: usize = 1 << 12;
+/// Default for the maximum number of bytes that can be received (16Kb).
+pub const DEFAULT_MAX_RECV_LIMIT: usize = 1 << 14;
```
th4s, line 8: What is the difference between these and `DEFAULT_TRANSCRIPT_MAX_SIZE` in `tls-mpc/src/config.rs`?

themighty1 (approved):
gw

PR #439, opened 2024-02-14
Show basic html info response for notary server's root endpoint

heeckhau:


notary-server/src/server.rs
```diff
@@ -111,7 +111,23 @@ pub async fn run_server(config: &NotaryServerProperties) -> Result<(), NotarySer
     let git_commit_hash = env!("GIT_COMMIT_HASH").to_string();
     let git_commit_timestamp = env!("GIT_COMMIT_TIMESTAMP").to_string();
 
+    let html_info = Html(format!(
+        r#"<h1>TLSNotary server {}!</h1>
```
yuroitaki, line 115:
```suggestion
        r#"<h1>TLSNotary Server {}!</h1>
```

notary-server/src/server.rs
```diff
@@ -111,7 +111,23 @@ pub async fn run_server(config: &NotaryServerProperties) -> Result<(), NotarySer
...
+        <ul>
+        <li>git commit hash: <a href="https://github.com/tlsnotary/tlsn/commit/{}">{}</a></li>
```
yuroitaki, line 117: I feel like it's bit redundant to have the git info and pub key repeated here — that it's sufficient to have just the health check and info hyperlinks — but wanna understand what's your motivation though of repeating them here?

yuroitaki (commented):
Great stuff! Just a few comments :)

notary-server/src/server.rs
heeckhau: I believe it's acceptable to repeat important information. :-)

Last week, there was a bit of confusion when someone reported that the notary was offline, which was not the case. This page primarily serves to demonstrate that the notary server is active when users quickly check the default URL. Therefore, while the actual content may not be critically important, I thought this info is helpful.

yuroitaki:
Almost forgot — we need to update the openapi.yaml to add this new homepage endpoint

heeckhau:
> Almost forgot — we need to update the openapi.yaml to add this new homepage endpoint

I don't consider this as API, this is intended for humans only :-)

yuroitaki (approved):
🚀

PR #447, opened 2024-02-26
Update verifier.rs to use web_time::Instant

maceip:
wasm-bindgen barfs on time


themighty1:
Hi, thanks, but that branch is still wip and is subject to changes. In the final version `time` will not be needed. lmk if i misunderstood your intent.

PR #448, opened 2024-02-26
Update prover.rs to use web_time::Instant

maceip:


themighty1:
Thanks, same as https://github.com/tlsnotary/tlsn/pull/447

PR #449, opened 2024-02-26
Update Cargo.toml to use web-time = "0.2"

maceip:


themighty1:
Thanks, same as https://github.com/tlsnotary/tlsn/pull/447

PR #456, opened 2024-03-14
Add example httpbin

th4s:
This PR adds an example which uses httpbin.


th4s:
I agree, I will turn this into an example.

th4s:
This needs more polish as it is turning into a refactor of examples.

tlsn/examples/src/lib.rs
```diff
@@ -155,3 +159,9 @@ pub async fn request_notarization(
         notarization_response.session_id,
     )
 }
+
+pub fn print_with_time(input: impl ToString, time: Instant) {
```
yuroitaki, line 163: can use the tracing crate for logging so verbosity level can be controlled easier :)

tlsn/examples/src/arg_parse.rs
```diff
@@ -0,0 +1,60 @@
+use std::collections::VecDeque;
```
yuroitaki, line 1: i use https://docs.rs/structopt/latest/structopt/ for CLI parsing, aha just suggesting in case you find it better, but feel free to stick to this :)

yuroitaki:
> This needs more polish as it is turning into a refactor of examples. Do not review yet.

oops sorry i just saw this 😅 , feel free to ignore my previous comments

th4s:
Closing because outdated.

PR #457, opened 2024-03-15
feat: automated network benches

sinui0:
This PR refactors our network bench suite, adding a script which automates it. It also now collects metrics and writes them to a file for easier visualization.

It also adds the ability to shutdown the server fixture using a query param, and makes our `tlsn-prover` and `tlsn-verifier` errors `Sync` as they should be.

### DIFF
--- a/tlsn/benches/src/lib.rs
+++ b/tlsn/benches/src/lib.rs
@@ -48,9 +48,8 @@ pub fn set_up() -> io::Result<()> {
 
 pub fn clean_up() {
     // Delete interface pair
-    if let Err(e) = Command::new("sudo")
+    if let Err(e) = Command::new("ip")
         .args(&[
-            "ip",
             "netns",
             "exec",
             PROVER_NAMESPACE,
@@ -65,15 +64,15 @@ pub fn clean_up() {
     }
 
     // Delete namespaces
-    if let Err(e) = Command::new("sudo")
-        .args(&["ip", "netns", "del", PROVER_NAMESPACE])
+    if let Err(e) = Command::new("ip")
+        .args(&["netns", "del", PROVER_NAMESPACE])
         .status()
     {
         println!("Error deleting namespace {}: {}", PROVER_NAMESPACE, e);
     }
 
-    if let Err(e) = Command::new("sudo")
-        .args(&["ip", "netns", "del", VERIFIER_NAMESPACE])
+    if let Err(e) = Command::new("ip")
+        .args(&["netns", "del", VERIFIER_NAMESPACE])
         .status()
     {
         println!("Error deleting namespace {}: {}", VERIFIER_NAMESPACE, e);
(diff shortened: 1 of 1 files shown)


tlsn/benches/bin/prover.rs
```diff
@@ -0,0 +1,183 @@
...
+            metric_wtr.serialize(
+                run_instance(instance, io)
+                    .await
+                    .context("failed to run instance")?,
+            )?;
+            metric_wtr.flush()?;
+        }
+    }
+
+    drop(metric_wtr);
```
yuroitaki, line 72: why do `metric_wtr` and `file` need to be manually dropped here? instead of automatically at the end of `main()`

tlsn/benches/README.md
```diff
@@ -4,118 +4,32 @@ This crate provides utilities for benchmarking protocol performance under variou
...
-```
-
-## Clean up
-
-For future reference, you can clean up this configuration as shown below.
-
-First, delete the interface pair (this removes both):
+## Preliminaries
 
+To run the benchmarks you will need `iproute2` installed, eg:
```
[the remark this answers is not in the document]
yuroitaki: ah sorry i didn't get to run it before this PR. From a brief search macos doesn't support `tc` and `ip`, probably need to make these swappable to macos compatible libs. Found a macos wrapper of iproute2 (https://github.com/brona/iproute2mac), but doesn't seem like it supports `tc` hmm

tlsn/benches/src/lib.rs
```diff
@@ -1,6 +1,259 @@
...
+        VERIFIER_INTERFACE,
+        VERIFIER_SUBNET.split('/').nth(0).unwrap(),
+    )?;
+
+    Ok(())
+}
+
+pub fn clean_up() {
+    // Delete interface pair
+    if let Err(e) = Command::new("sudo")
```
heeckhau, line 51: The bench is already running as root, so this (and the next) `sudo` are not necessary.

heeckhau (approved):
Works for me

tlsn/benches/src/metrics.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Debug, Clone, Serialize)]
+pub struct Metrics {
+    pub name: String,
+    pub upload: usize,
+    pub upload_delay: usize,
+    pub download: usize,
+    pub download_delay: usize,
+    pub upload_size: usize,
+    pub download_size: usize,
+    pub defer_decryption: bool,
```
themighty1, line 12:
```suggestion
    pub name: String,
    /// Upload bandwidth in Mbps.
    pub upload: usize,
    /// Upload latency in ms.
    pub upload_delay: usize,
    /// Download bandwidth in Mbps.
    pub download: usize,
    /// Download latency in ms.
    pub download_delay: usize,
    /// Total bytes sent to the server.
    pub upload_size: usize,
    /// Total bytes received from the server.
    pub download_size: usize,
    /// Whether deferred decryption was used.
    pub defer_decryption: bool,
```

themighty1 (approved):
gw, works for me.

PR #459, opened 2024-03-19
Deployment of multiple stable versions

ntampakas:
This PR is an enhancement of #419 and introduces the capability of running the three latest stable versions simultaneously on the system.
The requirement was to have all these versions deployed under the same EC2 instance.
The user can access a specific version, by adding it as a path at the end of notary URL. F.x.:
**https://notary.pse.dev/v0.1.0-alpha.1** or **https://notary.pse.dev/v0.1.0-alpha.2**, etc.

The deployment of the backend, is (still) being accomplished using AWS CodeDeploy. To keep the complexity of the appspec-scripts as simple as possible, it was decided to trigger the proxy using [AWS SSM](https://docs.aws.amazon.com/systems-manager/latest/userguide/what-is-systems-manager.html).
A new step was added to the workflow as well, which is executed only when a stable deployment is triggered. Its purpose is to make the appropriated changes to the proxy server, using the method stated above.

Additionally, some minor enhancements were applied, such as adding more filters when querying AWS resources, removing unused PATH, and changing shell options values.


scripts/modify_proxy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
yuroitaki, line 1: can we have a folder called `cd-scripts` and put this one inside, as well as the entire appspec-scripts folder?

appspec-scripts/start_app.sh
```diff
@@ -1,10 +1,25 @@
 #!/bin/bash
-set -e
-export PATH=$PATH:/home/ubuntu/.cargo/bin
+set -x
```
yuroitaki, line 2: we dont need -e for early termination anymore?

appspec-scripts/after_install.sh
```diff
@@ -1,20 +1,35 @@
...
-aws s3 cp s3://tlsn-deploy/$APP_NAME/.git ~/$APP_NAME/tlsn/.git --recursive
+if [ $APP_NAME = "stable" ]; then
+  # Prepare directories for stable versions
+  sudo mkdir ~/${APP_NAME}_${TAG}
+  sudo mv ~/tlsn ~/${APP_NAME}_${TAG}
+  sudo mkdir -p ~/${APP_NAME}_${TAG}/tlsn/notary-server/target/release
+  sudo chown -R ubuntu.ubuntu ~/${APP_NAME}_${TAG}
+
+  # Download .git directory
+  aws s3 cp s3://tlsn-deploy/$APP_NAME/.git ~/${APP_NAME}_${TAG}/tlsn/.git --recursive
```
yuroitaki, line 15: so s3 will always only contain the latest version of binary and .git right?

appspec-scripts/stop_app.sh
```diff
@@ -1,9 +1,35 @@
 #!/bin/bash
```
yuroitaki, line 1: can we put this link as a comment here as a reminder of the codedeploy hook sequence? https://docs.aws.amazon.com/codedeploy/latest/userguide/reference-appspec-file-structure-hooks.html#appspec-hooks-server

haha mainly for future reference to remind me and the team that stop_app always run first

appspec-scripts/start_app.sh
```diff
@@ -1,10 +1,25 @@
...
+  for PORT in $STABLE_PORTS; do
+    PORT_LISTENING=$(netstat -lnt4 | egrep -cw $PORT)
+    if [ $PORT_LISTENING -eq 0 ]; then
+      cd ~/${APP_NAME}_${TAG}/tlsn/notary-server
+      target/release/notary-server --config-file ~/.notary/${APP_NAME}_${PORT}/config.yaml &> ~/${APP_NAME}_${TAG}/tlsn/notary.log &
+      # Create a tag that will be used for service validation
+      INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id)
+      aws ec2 create-tags --resources $INSTANCE_ID --tags "Key=port,Value=$PORT"
+    break
+    fi
```
yuroitaki, line 18: if no port is available, shouldn't we also exit the script and fail the deployment since something must have gone wrong?

scripts/modify_proxy.sh
yuroitaki: also can we add a short comment saying that this script is to map port to the latest service?

yuroitaki (changes requested):
Amazing stuff <3 just a few minor comments and questions, most of them are to add comments so I can remember how this works in the future 😅

appspec-scripts/start_app.sh
ntampakas: **set -e** will make the script to exit immediately, if any command returns a non-zero exit code.
The way this script is implemented, defines a list of ports and loops through each one of them; then, it executes netstat to check if the port is in use. If not, **netstart.+egrep** command returns RC=1, causing the script to exit. This is why it was decided to eliminate this specific parameter.

yuroitaki:
Ah i guess one more question, in this setup ec2 instance tag is heavily used to pass the right version and port around, will there be any risk of these values being mistakenly modified by anyone on the AWS website? Was wondering if there is a safer place to store them, but if you or your team deem it's pretty low risk or there isn't better options, then I am totally fine with ec2 tag :)

appspec-scripts/after_install.sh
ntampakas: Correct! Since the workflow execution is performed sequentially, each new workflow execution, will upload the latest version to S3.

appspec-scripts/stop_app.sh
```diff
@@ -1,9 +1,35 @@
...
-set -e
+set -x
```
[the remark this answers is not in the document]
ntampakas: This parameter was removed, for the same reasons like mentioned here: https://github.com/tlsnotary/tlsn/pull/459#discussion_r1532049999
Additionally, CodeDeploy does not flag and stop the deployment in case of ApplicationStop hook failure.

cd-scripts/appspec-scripts/before_install.sh
```diff
@@ -3,8 +3,10 @@
```
yuroitaki, line 3: there's also a set -e in this script but commented out, is that also because of similar reason you described in other places?

appspec-scripts/start_app.sh
yuroitaki: i see, i'm just worried that an uncaught error from a command in these scripts eventually causes the service to be deployed wrongly; but since we have the validate_app check, it seems like it should be fine?

appspec-scripts/stop_app.sh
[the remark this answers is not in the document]
yuroitaki: i see, wouldn't it be dangerous then if codedeploy doesn't stop if stop_app hook fails to kill the old nightly service, and then start_app also fails to spin up the new nightly due to port still in use, and since start_app has exit 0 at the end, CD will continue, and validate_app will pass also since the old nightly is still running?

appspec-scripts/start_app.sh
yuroitaki: actually wouldn't the validation still pass? assuming that the instance port tag is always assigned with a value (since 3 versions are running since the beginning), and the oldest stable service fails to be killed (similar to  https://github.com/tlsnotary/tlsn/pull/459#discussion_r1533286846), then the validation check will still able to detect that a stable service is still listening at the port?

yuroitaki: maybe one way to solve this is in validation check, we ensure that version of the stable service that is running at the port matches the instance version tag (since that one is guaranteed to be true)

appspec-scripts/start_app.sh
ntampakas: > i see, i'm just worried that an uncaught error from a command in these scripts eventually causes the service to be deployed wrongly; but since we have the validate_app check, it seems like it should be fine?

Correct! :)

appspec-scripts/start_app.sh
yuroitaki: and for the nightly problem (https://github.com/tlsnotary/tlsn/pull/459#discussion_r1533286846), since we don't have version there, maybe validation needs to ensure that the nightly running at that port is not older than X hours/day?

appspec-scripts/stop_app.sh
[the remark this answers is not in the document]
ntampakas: I can understand your concern. Let me modify the scripts in order to support the **-e** argument :)

cd-scripts/appspec-scripts/before_install.sh
ntampakas: This will be fixed! Thank you!

appspec-scripts/start_app.sh
ntampakas: Please check: https://github.com/tlsnotary/tlsn/pull/459#discussion_r1533325517

yuroitaki (approved):
Thanks a lot for answering all my queries and addressing all my concerns, great work! @ntampakas

heeckhau (approved):
Thanks a lot

Issue #423, opened 2024-02-07
Implement authdecode protocol


themighty1:
The protocol component deals with messages passed between the Authdecode prover and the Verifier. 
The zk backend which performs zk proof generation and verification can be mocked.

Issue #342, opened 2023-09-20
feat: Support custom extensions in attestation


sinui0:
The Prover should be able to commit to additional data they want bound to the attestation. Their own public key for example.

samajammin:
Could someone help me understand what this enables?

PR #463, opened 2024-03-26
`tlsn-core` redesign

sinui0:
This PR redesigns `tlsn-core` with the following objectives in mind:

1. Update data structures to support additional commitment types such as simple hashes of the plaintext generated from authdecode #423 
    - We can add the poseidon hash algorithm in a follow up PR.
2. Encapsulate encoding commitment functionality
3. Fix inconsistent usage of hashing algorithms across the codebase
4. Support smart contracts verifiers
    - Keccak-256
    - Canonical serialization format which can be easily implemented in Solidity
6. Support arbitrary extra data fields bound to the attestation #342

### DIFF
--- a/tlsn/tlsn-core/src/attestation/proof.rs
+++ b/tlsn/tlsn-core/src/attestation/proof.rs
@@ -3,7 +3,7 @@ use tls_core::verify::WebPkiVerifier;
 
 use crate::{
     attestation::{AttestationBody, AttestationHeader, FieldKind},
-    conn::{ServerIdentity, ServerIdentityProof, ServerIdentityProofError},
+    conn::{default_cert_verifier, ServerIdentity, ServerIdentityProof, ServerIdentityProofError},
     merkle::MerkleProof,
     substring::{SubstringProof, SubstringProofError},
     PartialTranscript,
@@ -80,6 +80,20 @@ pub struct AttestationProofOutput {
 }
 
 impl AttestationProof {
+    /// Verifies the proof against the attestation header.
+    ///
+    /// Uses the default certificate verifier for the server identity proof.
+    ///
+    /// # Arguments
+    ///
+    /// * `header` - The header attested to by a Notary.
+    pub fn verify_with_default_cert_verifier(
+        self,
+        header: &AttestationHeader,
+    ) -> Result<AttestationProofOutput, AttestationProofError> {
+        self.verify(header, &default_cert_verifier())
+    }
+
     /// Verifies the proof against the attestation header.
     ///
     /// # Arguments
--- a/tlsn/tlsn-core/src/conn/mod.rs
+++ b/tlsn/tlsn-core/src/conn/mod.rs
@@ -357,3 +357,15 @@ impl CertificateSecrets {
         Some(alg.hash(&bytes))
     }
 }
+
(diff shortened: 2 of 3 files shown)


tlsn/tlsn-core/src/serialize.rs
```diff
@@ -0,0 +1,178 @@
...
+    encoding::EncodingLeaf,
+    hash::PlaintextHash,
+    transcript::{SliceIdx, SubsequenceIdx},
+};
+
+/// Canonical serialization of TLSNotary types.
+///
+/// This trait is used to serialize types into a canonical byte representation.
+///
+/// It is critical that the serialization is deterministic and unambiguous.
```
th4s, line 18: How do we ensure this? Is this ensured by carefully implementing the trait?  Just asking because we could enforce this through the type system, like using some length prefix encoding for the fields.

th4s (approved):
Nice! lgtm :rocket:

I did review from bird's eye view only for now, but I think the module structure looks good. Have 2 small comments.

th4s (commented):
I am still not completely through (`substring` and `encoding` missing), but I am too afraid that I run into some timeout and my temporary review is lost ^^.

Nice work :rocket: overall looking very good. When there was something obvious I marked it with `todo` so that we do not forget it. Some general things (which are mostly because I review so early):
- Since this is nearly a complete rewrite we could take the chance and adapt the comment style everywhere.
- I think we need more tests.
- (Error handling, panics and clippy)

Two things:
- Should we more strongly enforce the uniqueness of some of the `Secret` and `Field` variants in `AttestationFull` and `AttestationBody`.
- If we expect to add many more hash algorithms, the enum approach might get annoying to maintain.

tlsn/tlsn-core/src/encoding/proof.rs
```diff
@@ -0,0 +1,158 @@
...
+            // Make sure the amount of data being proved is bounded.
+            total_opened += opened_len as u128;
+            if total_opened > MAX_TOTAL_COMMITTED_DATA as u128 {
+                return Err(EncodingProofError::ExceededMaxData);
+            }
+
+            // Make sure the opening length matches the ranges length.
+            if seq.data.len() != opened_len {
+                return Err(EncodingProofError::Malformed);
+            }
```
th4s, line 118: Because of line 107, `seq.data.len() == opened_len` should always hold.

tlsn/tlsn-core/src/encoding/proof.rs
```diff
@@ -0,0 +1,158 @@
...
+        // Verify that the expected hashes are present in the merkle tree.
+        //
+        // This proves the Prover committed to the purported data prior to the encoder
+        // seed being revealed. Ergo, if the encodings are authentic then the purported
+        // data is authentic.
+        inclusion_proof
+            .verify(&commitment.root, &indices, &leafs)
+            .map_err(|_| EncodingProofError::MissingCommitment)?;
+
+        Ok(transcript)
```
th4s, line 156: There is still no `sent` and `received` data in here, but I think this is correct at this place?! Should we add this as a comment for the function signature?

th4s: Yeah, i think we can ignore it. I rechecked and it seems to be wrong.

PR #471, opened 2024-04-08
Readme for interactive example

heeckhau:


yuroitaki (commented):
Nice, we probably should also add more instruction on how to run this interactive example in `examples/interactive/README.md` too like what we did here — https://github.com/tlsnotary/tlsn/blob/main/tlsn/examples/simple/README.md#1-notarize-httpsexamplecom ?

PR #472, opened 2024-04-09
fix: interactive example

sinui0:
This PR fixes the interactive example to no longer manually close the TLS connection, and it enables deferred decryption.


yuroitaki (approved):
LGTM!

PR #477, opened 2024-04-26
docs: fix style in components (except tls)

themighty1:
This PR fixes the style in all crates under components (except the tls folder)

### DIFF
--- a/components/cipher/stream-cipher/src/lib.rs
+++ b/components/cipher/stream-cipher/src/lib.rs
@@ -1,14 +1,16 @@
 //! This crate provides a 2PC stream cipher implementation using a block cipher in counter mode.
 //!
-//! Each party plays a specific role, either the `StreamCipherLeader` or the `StreamCipherFollower`. Both parties
-//! work together to encrypt and decrypt messages using a shared key.
+//! Each party plays a specific role, either the `StreamCipherLeader` or the `StreamCipherFollower`.
+//! Both parties work together to encrypt and decrypt messages using a shared key.
 //!
 //! # Transcript
 //!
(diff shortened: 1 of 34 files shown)


components/cipher/stream-cipher/src/lib.rs
```diff
@@ -131,8 +133,8 @@ where
     ///
     /// # Arguments
     ///
-    /// * `explicit_nonce`: The explicit nonce to use for the keystream.
-    /// * `ciphertext`: The ciphertext to decrypt.
+    /// * `expli cit_nonce` - The explicit nonce to use for the keystream.
```
heeckhau, line 136:
```suggestion
    /// * `explicit_nonce` - The explicit nonce to use for the keystream.
```
(accidental space)

components/universal-hash/src/ghash/ghash_core/mod.rs
```diff
@@ -81,12 +85,12 @@ fn compute_new_add_shares(new_add_odd_shares: &[Gf2_128], add_shares: &mut Vec<G
         .iter()
         .zip((add_shares.len() + 1..).step_by(2))
     {
-        // `add_shares` always have an even number of shares so we simply add the next odd share
+        // `add_shares` always have an even number of shares so we simply add the next odd share.
         add_shares.push(*odd_share);
 
-        // now we need to compute the next even share and add it
+        // Now we need to compute the next even share and add it
         // note that the n-th index corresponds to the (n+1)-th power, e.g. add_shares[4]
```
heeckhau, line 92:
```suggestion
        // Now we need to compute the next even share and add it.
        // Note that the n-th index corresponds to the (n+1)-th power, e.g. add_shares[4]
```

heeckhau (approved):
lgtm

I suggested a few corrections

PR #486, opened 2024-05-14
Use cargo test in CI to run doc tests too

heeckhau:


themighty1 (approved):
ACK, but we should also remove `--tests` in
      - name: "Integration Test"
      - name: "Integration Test all features"
      
      otherwise `notary-server`s doctests won't get tested

heeckhau:
image
@themighty1 I rewrote more than I originally planned, but I think both the code and the output looks nice.
Can you do another review?

themighty1 (approved):
gw

PR #496, opened 2024-06-12
refactor(key-exchange): mpz upgrade

sinui0:
This PR updates our key exchange crate to use the new `mpz` API.

# Changes
- A bit of general clean up including making the error type opaque
- Moved the `point-addition` functionality into the key exchange crate. Having it in a dedicated crate was a bit overkill and burdensome
- Utilizes the preprocessing API from `mpz` to preprocess the key exchange as much as possible prior to the online phase. The follower now also immediately forwards their key during this to reduce latency.
- Uses the new `serio` IO abstraction
- Key generation now is handled in this crate and it ensures the key shares are dropped (zeroed) after used as inputs to the MPC.

### DIFF
--- a/components/key-exchange/src/exchange.rs
+++ b/components/key-exchange/src/exchange.rs
@@ -453,7 +453,7 @@ mod tests {
     use mpz_garble::protocol::deap::mock::{create_mock_deap_vm, MockFollower, MockLeader};
     use mpz_share_conversion::ideal::{ideal_share_converter, IdealShareConverter};
     use p256::{NonZeroScalar, PublicKey, SecretKey};
-    use rand_chacha::ChaCha20Rng;
+    use rand_chacha::ChaCha12Rng;
     use rand_core::SeedableRng;
     use serio::channel::MemoryDuplex;
 
@@ -503,7 +503,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_key_exchange() {
-        let mut rng = ChaCha20Rng::from_seed([0_u8; 32]);
+        let mut rng = ChaCha12Rng::from_seed([0_u8; 32]);
 
         let leader_private_key = SecretKey::random(&mut rng);
         let follower_private_key = SecretKey::random(&mut rng);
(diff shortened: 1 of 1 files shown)


components/key-exchange/src/exchange.rs
```diff
@@ -63,405 +76,429 @@ pub struct KeyExchangeCore<PS, PR, E> {
...
+            &mut self.converter_0,
+            &mut self.converter_1,
+            server_key,
+            private_key,
+        )
+        .await
     }
 
-    async fn compute_pms_for(
+    // Computes the PMS using both parties shares, performing an equality check
```
themighty1, line 132:
```suggestion
    // Computes the PMS using both parties' shares, performing an equality check
```

components/key-exchange/src/exchange.rs
```diff
@@ -63,405 +76,429 @@ pub struct KeyExchangeCore<PS, PR, E> {
...
+        let eq: [u8; 32] = self
+            .executor
+            .decode(&[eq])
+            .await?
+            .pop()
+            .expect("output 0 is eq")
+            .try_into()
+            .expect("eq is 32 bytes");
 
         // Eq should be all zeros if pms_1 == pms_2
```
themighty1, line 189:
```suggestion
        // Eq should be all zeros if pms_1 == pms_2.
```

components/key-exchange/src/exchange.rs
```diff
@@ -494,75 +531,85 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_compute_pms_share() {
+    async fn test_compute_pms() {
         let mut rng = ChaCha20Rng::from_seed([0_u8; 32]);
```
themighty1, line 535: ChaCha12Rng

themighty1 (approved):
gw!

components/key-exchange/src/exchange.rs
```diff
@@ -63,405 +76,429 @@ pub struct KeyExchangeCore<PS, PR, E> {
...
+            .pop()
+            .expect("output 0 is eq")
+            .try_into()
+            .expect("eq is 32 bytes");
 
         // Eq should be all zeros if pms_1 == pms_2
         if eq != [0u8; 32] {
-            return Err(KeyExchangeError::CheckFailed);
+            return Err(KeyExchangeError::new(
+                ErrorKind::ShareConversion,
```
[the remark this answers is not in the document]
th4s: Ahh I see, I would have considered it its own error kind like `ErrorKind::UnequalPms` to stress that this might be a cheating attempt. Yeah but I guess it works both ways.

PR #497, opened 2024-06-12
refactor(prf): mpz upgrade

sinui0:
This PR updates our `prf` crate to use the new `mpz` API.

# Changes
- A bit of general clean up including making the error type opaque
- Utilizes the preprocessing API from `mpz` to preprocess the prf as much as possible prior to the online phase.
- The PRF verify data was switched to being public, which should cut down on latency due to not needing OTPs.

### DIFF
--- a/components/prf/hmac-sha256/src/lib.rs
+++ b/components/prf/hmac-sha256/src/lib.rs
@@ -130,7 +130,7 @@ mod tests {
             .new_thread(follower_ctx_1, follower_ot_send_1, follower_ot_recv_1)
             .unwrap();
 
-        // Setup public PMS for testing
+        // Set up public PMS for testing.
         let leader_pms = leader_thread_0.new_public_input::<[u8; 32]>("pms").unwrap();
         let follower_pms = follower_thread_0
             .new_public_input::<[u8; 32]>("pms")


components/prf/hmac-sha256/src/prf.rs
```diff
@@ -157,36 +196,38 @@ where
...
-            std::mem::replace(&mut self.state, state::State::Error).try_into_server_finished()?;
+        let State::ServerFinished { hash_state, sf_vd } = self.state.take() else {
+            return Err(PrfError::state("PRF not in server finished state"));
+        };
 
         let circ = SERVER_VD_CIRC.get().expect("server vd circuit is set");
 
         if let Some(handshake_hash) = handshake_hash {
-            self.thread_1
+            self.thread_0
```
th4s, line 226: This is now `thread_0` instead of `thread_1` but it doesn't matter, right?

th4s (commented):
Nice :+1:

Shouldn't we adapt the `Prf` trait to reflect that `cfvd` and `sfvd` are no longer private?

components/prf/hmac-sha256/src/prf.rs
```diff
@@ -199,17 +240,17 @@ where
             .await?;
 
         let vd = if handshake_hash.is_some() {
-            let mut outputs = self.thread_1.decode_private(&[sf_vd.vd]).await?;
+            let mut outputs = self.thread_0.decode(&[sf_vd.vd]).await?;
```
themighty1, line 243: as above, we need to make sure that `compute_server_finished_vd_private/blind` is now called `compute_server_finished_vd`.

themighty1 (changes requested):
gw! some changes may be needed.

themighty1 (approved):

PR #499, opened 2024-06-12
fix(tlsn-server-fixture): update cert fixtures

sinui0:
This PR updates our expired cert fixtures, a temporary measure until we replace them with certs generated at runtime.


components/tls/tls-server-fixture/src/root_ca.der
yuroitaki, line 1: I used to be confused by this der file, whether it's the private key or the cert, would it be clearer if it's named `root_ca_cert.der`?

components/tls/tls-server-fixture/src/test_server.der
yuroitaki, line 1: same as https://github.com/tlsnotary/tlsn/pull/499/files#r1637442633

yuroitaki (commented):
LGTM, only one comment

th4s (approved):
lgtm

PR #504, opened 2024-06-17
refactor(aead): mpz upgrade

sinui0:
This PR updates the `aead` crate to use the new `mpz` API.

# Changes
- A bit of general clean up, changed to opaque error.
- Adds additional preprocessing
- Encapsulates the tag functionality a bit better.
- Removes commit-reveal from tag protocol for sent messages

### DIFF
--- a/components/aead/src/aes_gcm/tag.rs
+++ b/components/aead/src/aes_gcm/tag.rs
@@ -59,7 +59,7 @@ async fn compute_tag_share<C: StreamCipher<Aes128Ctr> + ?Sized, H: UniversalHash
 
 /// Computes the tag for a ciphertext and additional data.
 ///
-/// The commit reveal step is not required for computing a tag sent to the Server, as it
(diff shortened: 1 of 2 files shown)


th4s (approved):
:+1: gw

components/aead/src/aes_gcm/tag.rs
```diff
@@ -1,47 +1,160 @@
...
+
+    let tag = tag_share + other_tag_share;
+
+    Ok(tag)
+}
+
+/// Verifies a purported tag against the ciphertext and additional data.
+///
+/// Verifying a tag requires a commit reveal protocol between the leader and follower.
+/// Without it, the party which receives the others tag share first could trivially compute
```
themighty1, line 93:
```suggestion
/// Verifying a tag requires a commit-reveal protocol between the leader and follower.
/// Without it, the party which receives the other's tag share first could trivially compute
```

themighty1 (approved):
nice work. some non-blocking qs.

PR #505, opened 2024-06-17
refactor(prf): relax hiding of transcript hash, optimize latency

sinui0:
This PR refactors the `prf` crate with a focus on reducing online time. This should remove several communication rounds from the online phase as we no longer need to perform oblivious transfer as inputs become available.

# Changes
- Decoupled setup from preprocessing.
- Made handshake transcript hashes and `server_random` public inputs, keeping `client_random` private.
- Added a method to facilitate committing the `client_random` in the offline phase.

### DIFF
--- a/components/prf/hmac-sha256/src/prf.rs
+++ b/components/prf/hmac-sha256/src/prf.rs
@@ -235,14 +235,16 @@ where
         let thread = &mut self.thread_0;
 
         let randoms = Randoms {
+            // The client random is kept private so that the handshake transcript
+            // hashes do not leak information about the server's identity.
             client_random: thread.new_input::<[u8; 32]>(
-                "client_finished",
+                "client_random",
                 match self.config.role {
                     Role::Leader => Visibility::Private,
                     Role::Follower => Visibility::Blind,
                 },
             )?,
-            server_random: thread.new_input::<[u8; 32]>("server_finished", Visibility::Public)?,
+            server_random: thread.new_input::<[u8; 32]>("server_random", Visibility::Public)?,
         };
 
         let keys = SessionKeys {


components/prf/hmac-sha256/src/prf.rs
```diff
@@ -277,153 +279,163 @@ where
...
+
+            self.thread_0
+                .assign(&randoms.client_random, client_random)?;
+        } else if client_random.is_some() {
+            return Err(PrfError::role("only leader can set client random"));
+        }
+
+        self.thread_0
+            .commit(&[randoms.client_random.clone()])
+            .await?;
```
th4s, line 301: Why does this have to be called by the Follower?

th4s (approved):
:+1:

components/prf/hmac-sha256/src/prf.rs
```diff
@@ -246,24 +227,45 @@ where
...
         };
 
-        let visibility = match self.config.role {
-            Role::Leader => Visibility::Private,
-            Role::Follower => Visibility::Blind,
+        let thread = &mut self.thread_0;
+
+        let randoms = Randoms {
+            client_random: thread.new_input::<[u8; 32]>(
+                "client_finished",
```
themighty1, line 239:
```suggestion
                "client_random",
```

components/prf/hmac-sha256/src/prf.rs
```diff
@@ -246,24 +227,45 @@ where
...
+                match self.config.role {
+                    Role::Leader => Visibility::Private,
+                    Role::Follower => Visibility::Blind,
+                },
+            )?,
+            server_random: thread.new_input::<[u8; 32]>("server_finished", Visibility::Public)?,
```
themighty1, line 245:
```suggestion
            server_random: thread.new_input::<[u8; 32]>("server_random", Visibility::Public)?,
```

components/prf/hmac-sha256/src/prf.rs
```diff
@@ -246,24 +227,45 @@ where
...
-        let visibility = match self.config.role {
-            Role::Leader => Visibility::Private,
-            Role::Follower => Visibility::Blind,
+        let thread = &mut self.thread_0;
+
+        let randoms = Randoms {
+            client_random: thread.new_input::<[u8; 32]>(
+                "client_finished",
+                match self.config.role {
+                    Role::Leader => Visibility::Private,
```
themighty1, line 241: Can we add here an explanatory comment lest we forget. Why are we making the client_random private? I'm currently not sure myself.
Is it A) 
to add entropy to the handshake to enable domain name privacy during attestation (still need to investigate if server signature alone has enough entropy). If not, then is it B) to simplify the code logic and to avoid sending the `client_random` to the follower in the online phase?

themighty1 (approved):
gw!

PR #506, opened 2024-06-17
chore(uid-mux): delete crate, moved to tlsn-utils

sinui0:
This PR deletes the `uid-mux` crate which was moved to `tlsn-utils` repo.


th4s (commented):
Shouldn't we also update the `dependencies` section of crates depending on it?

themighty1 (approved):

PR #508, opened 2024-06-19
refactor: restructure notary crates

yuroitaki:
Relocate notary client crate into a notary folder that is restructured to contain the client, server and integration test crates — the downstream impacts are generally just
- examples
- CD and GH action


heeckhau (approved):
Looks good to me.

We'll have to update the [docs](https://docs.tlsnotary.org/developers/notary_server.html) too when we release

ntampakas (approved):
LGTM!

.github/workflows/cd-server.yml
```diff
@@ -49,8 +49,8 @@ jobs:
         uses: lewagon/wait-on-check-action@v1.3.1
         with:
           ref: ${{ github.ref }}
-          # Have to be specify '(notary-server)', as we are using matrix for build_and_test job in ci.yml, else it will fail, more details [here](https://github.com/lewagon/wait-on-check-action#check-name)
-          check-name: 'Build and test (notary-server)'
+          # Have to be specify '(notary)', as we are using matrix for build_and_test job in ci.yml, else it will fail, more details [here](https://github.com/lewagon/wait-on-check-action#check-name)
```
themighty1, line 52:
```suggestion
          # Have to specify '(notary)', as we are using matrix for build_and_test job in ci.yml, else it will fail, more details [here](https://github.com/lewagon/wait-on-check-action#check-name)
```

.github/workflows/cd.yml
```diff
@@ -20,8 +20,8 @@ jobs:
         uses: lewagon/wait-on-check-action@v1.3.1
         with:
           ref: ${{ github.ref }}
-          # Have to be specify '(notary-server)', as we are using matrix for build_and_test job in ci.yml, else it will fail, more details [here](https://github.com/lewagon/wait-on-check-action#check-name)
-          check-name: 'Build and test (notary-server)'
+          # Have to be specify '(notary)', as we are using matrix for build_and_test job in ci.yml, else it will fail, more details [here](https://github.com/lewagon/wait-on-check-action#check-name)
```
themighty1, line 23:
```suggestion
          # Have to specify '(notary)', as we are using matrix for build_and_test job in ci.yml, else it will fail, more details [here](https://github.com/lewagon/wait-on-check-action#check-name)
```

notary/server/README.md
```diff
@@ -109,7 +109,7 @@ To perform notarization, some parameters need to be configured by the prover and
 To streamline this process, a single HTTP endpoint (`/session`) is used by both TCP and WebSocket clients.
 
 #### Notarization
-After calling the configuration endpoint above, prover can proceed to start notarization. For TCP client, that means calling the `/notarize` endpoint using HTTP (`https`), while WebSocket client should call the same endpoint but using WebSocket (`wss`). Example implementations of these clients can be found in the [integration test](./tests/integration_test.rs).
+After calling the configuration endpoint above, prover can proceed to start notarization. For TCP client, that means calling the `/notarize` endpoint using HTTP (`https`), while WebSocket client should call the same endpoint but using WebSocket (`wss`). Example implementations of these clients can be found in the [integration test](../tests-integration/tests/notary.rs).
```
themighty1, line 112: I have a general comment that the "telegram style" which is acceptable when commenting the code where we omit the articles "a , the" is not appropriate imo for docs like this README. It seems like in this README there is a mixture of both style. I'll just fix this section to show what I mean (maybe someone go over our docs later and fix the rest)

```suggestion
After calling the configuration endpoint above, the prover can proceed to start the notarization. For a TCP client, that means calling the `/notarize` endpoint using HTTP (`https`), while a WebSocket client should call the same endpoint but using WebSocket (`wss`). Example implementations of these clients can be found in the [integration test](../tests-integration/tests/notary.rs).
```

themighty1 (approved):
lgtm, i mainly checked for typos

notary/server/README.md
yuroitaki: Good spot!

themighty1 (approved):

PR #512, opened 2024-06-21
refactor(tlsn): mpz upgrade

sinui0:
This PR updates `tlsn` crates to integrate the `mpz` upgrade.

*note: cherry pick #499 to run integration tests*

# Changes
1. Switch to using `serio` and new `uid-mux`, clean up IO code including better error handling. This should also remove any potential race conditions that may occur from the remote peer closing.
2. As changed in #502 we no longer reveal the MAC key to the verifier
3. `tracing` is always on
4. Added `force-st` feature to `tlsn-prover` and `tlsn-verifier` which forces single threaded execution (disables rayon)
5. We no longer allocate OTs for key exchange / GHASH upfront, it's handled downstream #496 #502 
6. Added some type aliases in `tlsn-common` to reduce noise

### DIFF
--- a/tlsn/tlsn-prover/src/tls/prove.rs
+++ b/tlsn/tlsn-prover/src/tls/prove.rs
@@ -163,6 +163,7 @@ impl Prover<ProveState> {
             })
             .await?;
 
+        // Wait for the verifier to correctly close the connection.
         if !mux_fut.is_complete() {
             mux_ctrl.mux().close();
             mux_fut.await?;
--- a/tlsn/tlsn-common/Cargo.toml
+++ b/tlsn/tlsn-common/Cargo.toml
@@ -8,12 +8,8 @@ edition = "2021"
 default = []
 
 [dependencies]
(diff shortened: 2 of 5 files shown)


themighty1:
@sinui0 , I think the conflicts need to be resolved. i cant load up this branch into vscode currently.

tlsn/tlsn-prover/src/tls/prove.rs
```diff
@@ -169,38 +146,28 @@ impl Prover<ProveState> {
...
+                // Send session_info to the verifier
+                io.send(session_info).await?;
+
+                Ok::<_, ProverError>(())
+            })
+            .await?;
+
+        if !mux_fut.is_complete() {
+            mux_ctrl.mux().close();
+            mux_fut.await?;
```
themighty1, line 168: need to add a comment

// Wait for the verifier to correctly close the connection.

themighty1 (approved):
nice work!!! i left some non-blocking qs

Issue #500, opened 2024-06-12
Add CI job to run expensive integration tests


sinui0:
Some integration tests, such as in `tls-mpc` are expensive so they are marked `ignore`, but these still need to be run in CI.

heeckhau:
@sinui0 Do you want an **extra CI step** to run all ignored tests?
```
cargo test -- --ignored
```
, we can also include the ignored tests when we run the other tests:
```
cargo test -- --include-ignored
```

We could also introduce a `expensive_tests` feature, if we need to run only a few specific tests and ignore other tests.

yuroitaki:
@sinui0 seems like this has been addressed by https://github.com/tlsnotary/tlsn/pull/537?

PR #514, opened 2024-06-24
chore: remove stale test and clean up ci job

sinui0:
This PR removes `components/integration-tests` as it is redundant to the integration test in `tls-mpc` and just adds maintenance burden. I also adjusted the CI to remove deleted packages and run `ignored` tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,7 @@ jobs:
       - name: "Test ${{ matrix.release && '--release' || '' }} ${{ matrix.all-features && '--all-features' || '' }}"
         env:
           RELEASE_OPTION: ${{ matrix.release && '--release' || '' }}
+          # Includes tests marked as `ignore`.
           IGNORED_OPTION: ${{ matrix.ignored && '--include-ignored' || '' }}
           ALL_FEATURES_OPTION: ${{ matrix.all-features && '--all-features' || '' }}
         # Run all tests (bins, examples, lib, integration and docs)


.github/workflows/ci.yml
```diff
@@ -34,7 +32,7 @@ jobs:
           - tlsn
           - notary-server
         include:
-          - package: components/integration-tests
+          - package: components/tls/tls-mpc
```
heeckhau, line 35: This will run the tests in `tls-mpc` twice, since they are already in `tls` above
See test run at https://github.com/tlsnotary/tlsn/actions/runs/9658293628/job/26639185063?pr=517

.github/workflows/ci.yml
```diff
@@ -81,7 +79,7 @@ jobs:
         # https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection
         run: |
           echo "Running command: cargo test $RELEASE_OPTION $ALL_FEATURES_OPTION"
-          cargo test $RELEASE_OPTION $ALL_FEATURES_OPTION
+          cargo test $RELEASE_OPTION $ALL_FEATURES_OPTION -- --include-ignored
```
heeckhau, line 82: This will always include the ignored tests. Maybe we can move this a few lines up inside the $RELEASE_OPTION ? https://github.com/tlsnotary/tlsn/issues/500

I'd also document this. suggestion: ```
        # `-- --include-ignored` to include the ignored (expensive integration) tests
```

heeckhau (changes requested):
I don't think these changes will do exactly what you intended

PR #516, opened 2024-06-25
Fix cd script.

yuroitaki:
Fix failed deployment: https://github.com/tlsnotary/tlsn/actions/runs/9655529116, where the `target/release/` folder is not in `notary/server` but `notary`

### DIFF
--- a/cd-scripts/appspec-scripts/stop_app.sh
+++ b/cd-scripts/appspec-scripts/stop_app.sh
@@ -24,7 +24,7 @@ if [ $APP_NAME = "stable" ]; then
     done
 
     echo "The oldest version is running under: $OLDEST_DIR"
-    PID=$(lsof $OLDEST_DIR/tlsn/notary/server/target/release/notary-server | awk '{ print $2 }' | tail -1)
+    PID=$(lsof $OLDEST_DIR/tlsn/notary/target/release/notary-server | awk '{ print $2 }' | tail -1)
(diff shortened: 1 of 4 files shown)


cd-scripts/appspec-scripts/stop_app.sh
```diff
@@ -24,7 +24,7 @@ if [ $APP_NAME = "stable" ]; then
     done
 
     echo "The oldest version is running under: $OLDEST_DIR"
-    PID=$(lsof $OLDEST_DIR/tlsn/notary/server/target/release/notary-server | awk '{ print $2 }' | tail -1)
+    PID=$(lsof $OLDEST_DIR/tlsn/notary/target/release/notary-server | awk '{ print $2 }' | tail -1)
```
yuroitaki, line 27: @ntampakas i realised this ll fail for the previous version folders that still have the old directory structure `.../notary-server/...`, i plan to manually update their directory structures and restart the services, so that future releases will be able to stop the old version app successfully, thoughts?

heeckhau (approved):

cd-scripts/appspec-scripts/stop_app.sh
ntampakas: It seems this is a really good idea! :+1:

ntampakas (approved):
LGTM!

Issue #518, opened 2024-06-25
Autogenerate self signed TLS certificates fixture


yuroitaki:
To autogenerate self signed TLS certificates fixture at runtime for testing purposes

yuroitaki:
### DRAFT

There are 3 places where these TLS fixtures are present:
(A) [tls-server-fixture](https://github.com/tlsnotary/tlsn/tree/main/components/tls/tls-server-fixture/src) — consumed by notary-server [integration test](https://github.com/tlsnotary/tlsn/blob/main/notary/integration-tests/tests/notary.rs)
(B) [tlsn-server-fixture](https://github.com/tlsnotary/tlsn/tree/main/tlsn/tlsn-server-fixture/src/tls) — consumed by tlsn [tests-integration](https://github.com/tlsnotary/tlsn/tree/main/tlsn/tests-integration/tests) and [bench](https://github.com/tlsnotary/tlsn/tree/main/tlsn/benches/bin)
(C) [notary-server](https://github.com/tlsnotary/tlsn/tree/main/notary/server/fixture/tls) — consumed by its [integration test
](https://github.com/tlsnotary/tlsn/blob/main/notary/integration-tests/tests/notary.rs)

My draft proposal for the implementation:
(A), (B) — Expose a function in these server fixtures that allow external crate (test or bench) to call. This function will use rust [openssl](https://docs.rs/openssl/latest/openssl/index.html) to generate all the required fixtures (rootCA key, cert, server key, server cert), store the server cert and key in memory so that the [bind](https://github.com/tlsnotary/tlsn/blob/main/tlsn/tlsn-server-fixture/src/lib.rs#L51) function can use them later, and return rootCA cert to the caller.
(C) — Generate the required fixtures in the integration test [itself](https://github.com/tlsnotary/tlsn/blob/main/notary/integration-tests/tests/notary.rs#L68), save the server key and cert to a temporary file location (to be deleted once tests are done) and pass their file paths to the [config](https://github.com/tlsnotary/tlsn/blob/main/notary/integration-tests/tests/notary.rs#L50-L51), and store the rootCA cert in memory to be used by the [tests](https://github.com/tlsnotary/tlsn/blob/main/notary/integration-tests/tests/notary.rs#L120) later.

As long as we use the vendored version of [openssl](https://docs.rs/openssl/latest/openssl/index.html#vendored), this should be able to run in any machine including GH action runner.

[...]

PR #519, opened 2024-06-25
Chore/update notary cert

yuroitaki:
Temporarily update self signed cert fixtures used in notary server before https://github.com/tlsnotary/tlsn/issues/518 is addressed


heeckhau (approved):
👍

PR #520, opened 2024-06-25
Update CD script for easier investigation

yuroitaki:
With this, one can just do `pgrep -af notary-server` to find out which version is running at which port using which config file

```bash
228356 /home/ubuntu/stable_v0.1.0-alpha.3/tlsn/notary/target/release/notary-server --config-file /home/ubuntu/.notary/stable_7047/config.yaml
```

### DIFF
--- a/cd-scripts/appspec-scripts/start_app.sh
+++ b/cd-scripts/appspec-scripts/start_app.sh
@@ -12,8 +12,7 @@ if [ $APP_NAME = "stable" ]; then
   for PORT in $STABLE_PORTS; do
     PORT_LISTENING=$(netstat -lnt4 | egrep -cw $PORT || true)
     if [ $PORT_LISTENING -eq 0 ]; then
-      cd ~/${APP_NAME}_${TAG}/tlsn/notary
-      target/release/notary-server --config-file ~/.notary/${APP_NAME}_${PORT}/config.yaml &> ~/${APP_NAME}_${TAG}/tlsn/notary.log &
+      ~/${APP_NAME}_${TAG}/tlsn/notary/target/release/notary-server --config-file ~/.notary/${APP_NAME}_${PORT}/config.yaml &> ~/${APP_NAME}_${TAG}/tlsn/notary.log &
       # Create a tag that will be used for service validation
       INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id)
       aws ec2 create-tags --resources $INSTANCE_ID --tags "Key=port,Value=$PORT"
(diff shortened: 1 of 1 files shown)


yuroitaki:
It's also to find out if a specific version is still running — currently the only reliable way I know to find out that (and which version is running at which port using which config file) is to inspect the corresponding log file

yuroitaki:
> It's also to find out if a specific version is still running — currently the only reliable way I know to find out that (and which version is running at which port using which config file) is to inspect the corresponding log file

@ntampakas let me know what you think

ntampakas:
Hey @yuroitaki ,
Your idea sounds really good.
Something like this could probably help us match the running stable version with the port:

```
for port in 7047 7057 7067; do
  netstat -lnt4 | egrep -cw $port || continue
  for version in $(find . -type d -name 'stable_v*'); do
    egrep -cw $port $version/tlsn/notary.log >/dev/null
    [ $? -ne 0 ] || echo "Version $version is runnind on port $port"
  done
done
```

yuroitaki:
> Hey @yuroitaki , Your idea sounds really good. Something like this could probably help us match the running stable version with the port:
> 
> ```
> for port in 7047 7057 7067; do
>   netstat -lnt4 | egrep -cw $port || continue
>   for version in $(find . -type d -name 'stable_v*'); do
>     egrep -cw $port $version/tlsn/notary.log >/dev/null
>     [ $? -ne 0 ] || echo "Version $version is runnind on port $port"
>   done
> done
> ```

thanks for the suggestion! in this case we need to check AWS CloudDeploy log to see this echo log?

ntampakas:
> > Hey @yuroitaki , Your idea sounds really good. Something like this could probably help us match the running stable version with the port:
> > ```
> > for port in 7047 7057 7067; do
> >   netstat -lnt4 | egrep -cw $port || continue
> >   for version in $(find . -type d -name 'stable_v*'); do
> >     egrep -cw $port $version/tlsn/notary.log >/dev/null
> >     [ $? -ne 0 ] || echo "Version $version is runnind on port $port"
> >   done
> > done
> > ```
> 
> thanks for the suggestion! in this case we need to check AWS CloudDeploy log to see this echo log?

Probably it doesn' t make sense to include something like this in CI/CD (and check the CloudDeploy logs).
The script above was mentioned as an _aux tool_ in a troubleshooting scenario on the EC2, just to quickly identify the listening ports for all stable versions running.
Additionally, since we use Grafana for log aggregation, you could use something like this to retrieve the running port for a specific version:
```
{filename="/home/ubuntu/stable_v0.1.0-alpha.6/tlsn/notary.log"} |= `Listening for TCP traffic`
```

Please let me know what you think.

ntampakas (approved):
LGTM!

heeckhau (approved):

PR #532, opened 2024-07-05
CI/CD check is added to identify the number of stable versions deployed

ntampakas:
An extra check was added to the CI/CD CodeDeploy scripts to verify if there are more than three stable versions deployed. If this condition is met, the deployment will be interrupted.


yuroitaki (commented):
Nice stuff! Just one question: is the reason why we have to put this logic here instead of `stop_app` that `exit 1` doesn't work well in `stop_app`?

yuroitaki (approved):

PR #537, opened 2024-07-22
refactor: consolidate workspaces

sinui0:
This PR consolidates all our crates into a single workspace. This should simplify our CI, improve compile times and reduce memory footprint of rust analyzer and speed up its first pass.

Dependency declarations were normalized to always use the fully qualified syntax:

```toml
[dependencies]
package = { version = "x" }
```
This is cleaner as it allows adding extra fields easily, eg enabling features.

If you experience rust analyzer being sluggish, you can configure it to only analyze specific packages using:

```json
"rust-analyzer.linkedProjects": [
  "crates/.."
]
```

### DIFF
--- a/crates/components/aead/src/aes_gcm/mod.rs
+++ b/crates/components/aead/src/aes_gcm/mod.rs
@@ -515,6 +515,7 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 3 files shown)


th4s (approved):
:rocket:

PR #538, opened 2024-07-23
fix: notary cd

sinui0:
This PR fixes the notary server CD after being broken by #537


th4s (approved):

heeckhau (approved):
Looking good, except for the failing test

PR #539, opened 2024-07-24
Enable deferred-decryption per default

th4s:
This PR enables deferred decryption per default. It exposes more fine-grained control over what size is decrypted locally after the TLS connection and what is decrypted online.

### DIFF
--- a/crates/prover/src/tls/config.rs
+++ b/crates/prover/src/tls/config.rs
@@ -21,9 +21,22 @@ pub struct ProverConfig {
     /// Maximum number of bytes that can be sent.
     #[builder(default = "DEFAULT_MAX_SENT_LIMIT")]
     max_sent_data: usize,
-    /// Maximum number of bytes that can be received.
+    /// Maximum number of bytes that can be decrypted online.
+    #[builder(default = "0")]
+    max_recv_data_online: usize,
+    /// Maximum number of bytes that will be decrypted after the TLS connection is closed.
     #[builder(default = "DEFAULT_MAX_RECV_LIMIT")]
-    max_recv_data: usize,
+    max_deferred_size: usize,
+    /// Defers the decryption from the start of the connection until after the connection is closed.
+    ///
+    /// Decryption of responses will be deferred until after the TLS connection is closed. This is
+    /// useful if you either have only one request/response cycle of if you have several such
+    /// cycles but the content of the request never depends on the content of the previous response.
+    ///
+    /// This allows to decrypt responses locally without MPC, so this option saves bandwidth
+    /// and performance.
+    #[builder(default = "true")]
+    defer_decryption_from_start: bool,
 }
 
 impl ProverConfig {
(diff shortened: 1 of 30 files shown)


crates/prover/src/tls/config.rs
```diff
@@ -100,11 +109,19 @@ impl ProverConfig {
     }
 
     pub(crate) fn ot_sender_setup_count(&self) -> usize {
-        ot_send_estimate(Role::Prover, self.max_sent_data, self.max_recv_data)
+        ot_send_estimate(
```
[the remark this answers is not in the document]
th4s: yeah I think this would make sense, 2 questions:

1. Why is `EXTRA_OTS = 16384`? Is this the value for key exchange, prf and ghash?
2. Why is `ot_send_estimate` for `Prover` only `EXTRA_OTS`? Same for `ot_recv_estimate` and `Verifier`?

crates/prover/src/tls/config.rs
```diff
@@ -22,11 +22,21 @@ pub struct ProverConfig {
...
     max_sent_data: usize,
     /// Maximum number of bytes that can be decrypted online.
-    #[builder(default = "DEFAULT_MAX_RECV_LIMIT")]
+    #[builder(default = "0")]
     max_recv_data_online: usize,
     /// Maximum number of bytes that will be decrypted after the TLS connection is closed.
-    #[builder(default = "0")]
+    #[builder(default = "DEFAULT_MAX_RECV_LIMIT")]
     max_deferred_size: usize,
+    /// Defers the decryption from the start of the connection.
```
heeckhau, line 30:
```suggestion
    /// Defers decryption until after the connection is closed.
```

crates/prover/src/tls/config.rs
```diff
@@ -22,11 +22,21 @@ pub struct ProverConfig {
...
+    ///
+    /// Decryption of responses will be deferred until after the TLS connection is closed. This is
+    /// useful if you either have only one request/response cycle of if you have several such
+    /// cycles but the content of the request never depends on the content of the previous response.
+    ///
+    /// This allows to decrypt responses locally without MPC, so this option saves bandwidth
+    /// and performance.
+    #[builder(default = "true")]
+    defer_decryption_from_start: bool,
```
heeckhau, line 39: I feel it would be better to refer to when it is deferred to (after close), instead of saying it is deferred at the start

th4s: Hmm I explicitly added `from_start` because we can also enable it midflight, so that people do not wonder what this is for when we also have `defer_decryption()`.

th4s:
Currently experiencing #559 locally...

th4s:
> Currently experiencing #559 locally when testing this branch.

This did not have anything to do with #559 and was a bug I introduced, but is now fixed. Rdy for review.

th4s:
The rebase is turning into a rewrite so it is easier to open a new PR and redo this.

PR #547, opened 2024-07-29
Comment failing tests in aead

th4s:
This PR comments the aead tests which currently time out.

### DIFF
--- a/crates/components/aead/src/aes_gcm/mod.rs
+++ b/crates/components/aead/src/aes_gcm/mod.rs
@@ -514,201 +514,203 @@ mod tests {
         (leader, follower)
     }
 
-    #[tokio::test]
-    #[ignore = "expensive"]
-    async fn test_aes_gcm_encrypt_private() {
-        let key = vec![0u8; 16];
-        let iv = vec![0u8; 4];
-        let explicit_nonce = vec![0u8; 8];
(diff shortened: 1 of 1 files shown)


heeckhau:
@th4s on my machine this test also fails:
https://github.com/tlsnotary/tlsn/blob/7451ecba92fee09d5df30142cee479f92101964d/crates/tls/mpc/tests/test.rs#L322

We should also rename this test. Test "test" was tricky to find :-)

heeckhau:
> We should also rename this test. Test "test" was tricky to find :-)

After I rename this test to `test_mpc_tls_common_config` it succeeds. Race?

heeckhau:
> > @th4s on my machine this test also hangs:
> > https://github.com/tlsnotary/tlsn/blob/7451ecba92fee09d5df30142cee479f92101964d/crates/tls/mpc/tests/test.rs#L322
> > 
> > We should also rename this test. Test "test" was tricky to find :-)
> 
> Youre the only one able to repro this locally! Can you please capture a trace log with NEW and CLOSE tracing events enabled?

I can't reproduce it anymore... all tests now succeed, even on `dev`

th4s:
Hmm should we then proceed and merge this PR if there are other tests that randomly fail?

PR #549, opened 2024-07-30
GH actions cleanup

heeckhau:


README.md
```diff
@@ -52,6 +52,27 @@ at your option.
...
+> ```
+
+If you run into this error:
+```
+Could not find directory of OpenSSL installation, and this `-sys` crate cannot
+  proceed without this knowledge. If OpenSSL is installed and this crate had
+  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
+  compilation process.
+```
+Make sure you have the development packages of OpenSSL installed (`libssl-dev` on Ubuntu or `openssl-devel` on Fedora).
```
yuroitaki, line 72: What about MacOS users?

README.md
```diff
@@ -52,6 +52,27 @@ at your option.
...
+
+If you want to contribute; Thanks a lot! please check <CONTRIBUTING.md> first.
```
yuroitaki, line 74:
```suggestion
If you want to contribute, thanks a lot! Please check [CONTRIBUTING.md](CONTRIBUTING.md) first.
```

crates/wasm/build.sh
```diff
@@ -0,0 +1,3 @@
+RUSTFLAGS='-C target-feature=+atomics,+bulk-memory,+mutable-globals' \
+    rustup run nightly \
+    wasm-pack build --target web . -- -Zbuild-std=panic_abort,std
```
yuroitaki, line 3: The command in https://github.com/tlsnotary/tlsn/blob/dev/crates/wasm/README.md doesn't have `--`, should we stick to one?

yuroitaki (commented):
Looks good~ just a few comments

README.md
heeckhau: I'll drop this line. There is Contribution section below it

README.md
yuroitaki: P/S: Doesn't seem to affect MacOS users

yuroitaki (approved):

themighty1 (approved):

PR #560, opened 2024-08-07
Add `verify` for wasm `TlsProof`

th4s:
This PR adds a `verify` method for `TlsProof` in wasm.


th4s:
@sinui0 I realize now that consuming `self` in `verify` might be suboptimal considering the wasm/js boundary. What do you think?

PR #561, opened 2024-08-09
refactor: use public key pem instead of bytes

0xtsukino:
@sinui0 @th4s what do you think about passing in the pem string directly to the verify method?

I've tried lots of different variation of decoding pem into byte array on JS side, but all of them failed and gave me the error `Invalid public key` in wasm 😞 

For reference, i've tried:
- [ec-key](https://www.npmjs.com/package/ec-key#-tobuffer-format-)
- [elliptic](https://www.npmjs.com/package/elliptic)
- decoding pem to base64 then convert to bytes
- decoding pem to base64, then encode to hex, then convert to bytes...


th4s (commented):
:+1:

PR #568, opened 2024-08-15
Fix hanging tls-mpc integration test

### LINKED ISSUE #559: Investigate hanging `tlsn/crates/tls/mpc/tests/test.rs`
When running the [tls-mpc-test](https://github.com/tlsnotary/tlsn/blob/dev/crates/tls/mpc/tests/test.rs) with a tracing configuration that emits `New` and `Close` [FmtSpans](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.FmtSpan.html) we were able to capture the following log [trace_hang.txt](https://github.com/user-attachments/files/16509460/trace_hang_trimmed.txt).

After filtering for lines that contain `decode_key_private` and writing all these `Leader` events into this file
[decode_key_private.txt](https://github.com/user-attachments/files/16509459/decode_key_01_private.txt) and doing the same for `decode_key_blind` and writing these `Follower` events into this file [decode_key_blind.txt](https://github.com/user-attachments/files/16509458/decode_key_01_blind.txt) I notice that `decode_key_private` is called twice on the `Leader` site (line 1 and 21). This is not true for the follower, where it is called only once (line 1). This is what causes the leader site to be hanging in `ot_receive_active_encodings` (line 14389 in `trace_hang.txt`).

The cause for this might be that [commit](https://github.com/tlsnotary/tlsn/blob/dev/crates/tls/mpc/src/leader.rs#L334) which calls `decode_key_private`is called twice:
- https://github.com/tlsnotary/tlsn/blob/dev/crates/tls/client-async/src/lib.rs#L179
- https://github.com/tlsnotary/tlsn/blob/dev/crates/tls/mpc/tests/test.rs#L194

th4s:
This PR fixes the hanging integration test and closes #559 by checking if the leader already committed.


th4s:
@heeckhau Can you check if this fixes your hanging test. If I remember correctly you were the only one who encountered this and provided the logs for #559.

heeckhau:
> @heeckhau Can you check if this fixes your hanging test. If I remember correctly you were the only one who encountered this and provided the logs for #559.

I ran it 3 times. 3 times without problems. It was often flaky before, so no guarantees

PR #570, opened 2024-08-19
test: add browser benches

themighty1:
This PR adds the functionality to benchmark the prover running in a browser.

- `binary` contains the binaries and the native bench  
- `library` contains logic shared by the native and the browser bench
- `browser` contains the `native` module which drives the browser and loads the wasm module.

The docker image was tested only on Linux.

The new github workflow was tested here https://github.com/themighty1/tlsn/actions/workflows/bench-browser.yml


.github/workflows/bench-browser.yml
```diff
@@ -0,0 +1,27 @@
+name: Run Benchmarks In The Browser
+on:
+  # manual trigger
```
heeckhau, line 3: @themighty1 To test this, I'd also add `pull_request:` here. Once we see it successfully runs on GH we can remove `pull_request:` again.

themighty1: Thanks, Im testing it in another repo https://github.com/themighty1/tlsn/actions

themighty1:
It appears that the bench would hang occasionally. I'll have to investigate it and then we can proceed with review.

themighty1:
The PR is ready for review and merging. Now with the `.incognito()` flag, the test suite is much more stable. 
I ran the suite in a loop 200 times and it did hang once. Before adding `incognito()`, it would hang on every 2nd run of the bench suite. 
It is hard to debug the reason for hanging because one needs to run the suite in a loop for 24-48 hours before the hang occurs. When we get our upcoming prover speed improvements, it will be easier to repro this issue.

Issue #548, opened 2024-07-29
Inspect timing-out aead tests


th4s:
PR #547 comments the aead tests, which are currently timing out. This needs further investigation as it is not locally reproducible.

th4s:
PR #558 is used for investigating.

th4s:
This deadlock is hard to fix. Things I tried so far:

- I could not reproduce this bug on my local machine, even when using act and running it in docker.
- Deadlock seems to vanish when executing a single AEAD test. It is strange that either all 4 of the started aead tests hang or all pass. I never observed some test passing and another hanging.
- Looking for `static` variables, i.e. shared state between unit tests -> nothing found
- Manual code inspection, especially of `mpz-garble/deap`-> nothing found so far. 
- Replacing all `.lock().unwrap()` with `try_lock().unwrap()` -> deadlock vanishes.
- Adding tracing.
  - There seems to be a problem with [`set_default`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html#method.set_default) in that it omits some logging statements. Couldn't make any sense out of it, will probably create an issue for this. However sometimes I was able to capture a little bit of output. This can be found in the failed runs in [these PRs](https://github.com/th4s/tlsn/pulls)
  - When trying to add proper tracing via [`set_global_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html) the deadlock vanishes.
- Removing tracing and using `println!` statements the deadlock vanishes.

Saying the deadlock vanishes means that I tried to run it several times (4-5) and didn't encounter it.

th4s:
Making some progress by carefully adding `println!` statements. This run is interesting: https://github.com/th4s/tlsn/actions/runs/10506339441/job/29106915720
It hangs in `decrypt_private -> verify_tag -> compute_tag_share -> share_keystream_block -> compute_keystream -> compute -> execute`

th4s:
New hanging run with more logging: https://github.com/th4s/tlsn/actions/runs/10507218244/job/29108495136
It hangs inside `generate` (also in `evaluate` but this is probably just a consequence of waiting for `generate`).

My current bet would be that it has to do with `feed, flush` and the limited backpressure of the `TestSTExecutor`.

th4s:
New hanging run with more fine-grained logging: https://github.com/th4s/tlsn/actions/runs/10509969553/job/29118455831
 
[...]

PR #575, opened 2024-08-27
Fix aead deadlock in unit tests

th4s:
This PR fixes the deadlock in our aead unit tests and adds the tests back. More info here #548.

### DIFF
--- a/crates/components/aead/src/aes_gcm/mod.rs
+++ b/crates/components/aead/src/aes_gcm/mod.rs
@@ -423,294 +423,306 @@ impl<Ctx: Context> Aead for MpcAesGcm<Ctx> {
 
 #[cfg(test)]
 mod tests {
-    // TODO: The following tests time-out on Github CI pipeline, which is not reproducible local.
-    // Needs further investigation.
+    use super::*;
+    use crate::{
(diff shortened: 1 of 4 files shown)


crates/components/aead/src/aes_gcm/mod.rs
```diff
@@ -423,294 +423,306 @@ impl<Ctx: Context> Aead for MpcAesGcm<Ctx> {
...
+    async fn setup_pair(
+        key: Vec<u8>,
+        iv: Vec<u8>,
+    ) -> (
+        MpcAesGcm<STExecutor<MemoryDuplex>>,
+        MpcAesGcm<STExecutor<MemoryDuplex>>,
+    ) {
+        INIT_RAYON_THREADPOOL.call_once(|| {
+            rayon::ThreadPoolBuilder::new()
+                .num_threads(9)
```
heeckhau, line 473: This is the actual fix. Maybe it is worth converting 9 into a CONSTANT and documenting in this file why you picked 9 as magic number.

th4s: I will wait a little bit, because maybe the fix is to ditch rayon, instead. We will discuss today.

heeckhau (approved):
looks good to me

themighty1 (approved):

PR #492, opened 2024-06-05
Notary Server SGX Attestation Template

maceip:
<h2>sgx attestation</h2>
this pr adds two files to the notary-server:

1) a gramine manifest template
   - this manifest details needed files, mounts, and other hardware required by the enclave; it sets the entrypoint to be executed when the enclave boots, and other gramine specific security / policy settings

2) a Makefile
   - a convenience file to preprocess the manifest template, and runs our binary in SGX, or if there is no SGX hardware, uses gramine-direct for testing / verification.

_this PR sets the stage for SGX specific concepts and scaffolding before we deploy on SGX hardware_

<h3><a href='https://gramine.readthedocs.io/en/stable/sgx-intro.html#term-SGX-Report'>The SGX Report</a></h3>

The SGX report contains a few values, one of which is  [MR_ENCLAVE](https://www.intel.com/content/www/us/en/developer/articles/technical/introduction-to-intel-sgx-sealing.html). 

This PR also adds a push-hook that generates a new SGX report for that commit, and modifies the project's README.md with the updated SGX Report:

<p align="center">
auto-report
</p>

#### How To Create an SGX Report without SGX:

_steps to test this out (gramine is a pain to build on macos, use a *nix:_

1. install gramine; for ubuntu it's (copypasta this whole block):
    
    - ```
      sudo curl -fsSLo /usr/share/keyrings/gramine-keyring.gpg https://packages.gramineproject.io/gramine-keyring.gpg
      echo "deb [arch=amd64 signed-by=/usr/share/keyrings/gramine-keyring.gpg] https://packages.gramineproject.io/ $(lsb_release -sc) main" \
      | sudo tee /etc/apt/sources.list.d/gramine.list
      sudo curl -fsSLo /usr/share/keyrings/intel-sgx-deb.asc https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key
      echo "deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx-deb.asc] https://download.01.org/intel-sgx/sgx_repo/ubuntu $(lsb_release -sc) main" \
      | sudo tee /etc/apt/sources.list.d/intel-sgx.list

      sudo apt-get update
      sudo apt-get install gramine
      ```

2. ensure you have the tlsn deps
3. ```make```
4. ```gramine-sgx-gen-private-key```
5. ```gramine-sgx-sign -v --manifest notary-server.manifest --output notary-server.sgx```
[...]

PR #580, opened 2024-09-11
ci: azure ssh runner for named feature: tee_quote 

maceip:


yuroitaki (changes requested):
Nice stuff! Can we combine this with https://github.com/tlsnotary/tlsn/pull/492 instead, i.e. in the same github workflow file? 

Cause since this build workflow is temporary (until devops set up azure), we can have it parked under there (.github/workflows/gramine-report.yml), where a push/pull request to the 'attest' branch will trigger 

(1) this build to run notary-server in azure
(2) calculating gramine measurement and store it in GH artefact

what do you think?

.github/workflows/attestation-build.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Checkout code
+      uses: actions/checkout@v4
+    - name: azure-demo ${{ github.sha }}
+      uses: appleboy/ssh-action@v1.0.3
+      with:
+         host: ${{ secrets.HOST }}
+         username: ${{ secrets.USERNAME }}
+         key: ${{ secrets.KEY }}
+         port: ${{ secrets.PORT }}
+         command_timeout: 120m
```
heeckhau, line 25: How long do you expect this to take?

.github/workflows/attestation-build.yml
```diff
@@ -0,0 +1,33 @@
...
+         script: |
+            cd /tmp 
+            git clone https://github.com/tlsnotary/tlsn ${{ github.sha }}
+            cd ${{ github.sha }}
+            git checkout ${{ github.sha }}
+            cd crates/notary/server
+            timeout 120 sh -c 'until nc -z $0 $1; do sleep 1; done' localhost 7047 &
```
heeckhau, line 32: Add a comment here: "Wait for port 7047 to be available before we start the notary" or something

.github/workflows/attestation-build.yml
```diff
@@ -0,0 +1,33 @@
...
+            /home/ubuntu/.cargo/bin/cargo run --bin notary-server --features attestation
```
heeckhau, line 33: * `cargo` is not on the $PATH?
* shouldn't this run in  `--release` ?

heeckhau (changes requested):
The base branch is wrong. Shouldn't this be based on #588 

And shouldn't `attestation` be renamed in `tee_quote`?

maceip:
> Nice stuff! Can we combine this with #492 instead, i.e. in the same github workflow file?
> 
> Cause since this build workflow is temporary (until devops set up azure), we can have it parked under there (.github/workflows/gramine-report.yml), where a push/pull request to the 'attest' branch will trigger
> 
> (1) this build to run notary-server in azure (2) calculating gramine measurement and store it in GH artefact
> 
> what do you think?

since this is using ssh auth id prefer to keep this seperate and simple => once we get the blessed azure auth i will rm this

.github/workflows/attestation-build.yml
maceip: better to be explicit -- since gh runners / action scripts are opaque, & this is a tmp setup, once we are properly linked to azure we will either use the rust container or build it every time and set PATH

yuroitaki:
> Cause since this build workflow is temporary (until devops set up azure), we can have it parked under there (.github/workflows/gramine-report.yml), where a push/pull request to the 'attest' branch will trigger

Alright cool, but we will need https://github.com/tlsnotary/tlsn/pull/492 to complete the GH artefact task right?

.github/workflows/azure-ssh-attestation-build.yml
```diff
@@ -0,0 +1,34 @@
...
+         command_timeout: 120m
+         script: |
+            cd /tmp 
+            git clone https://github.com/tlsnotary/tlsn ${{ github.sha }}
+            cd ${{ github.sha }}
+            git checkout ${{ github.sha }}
+            cd crates/notary/server
+            : # if nc cant dial 7047 within 120m: ❌; else ✅ 
+            timeout 120 sh -c 'until nc -z $0 $1; do sleep 1; done' localhost 7047 &
+            /home/ubuntu/.cargo/bin/cargo run --bin notary-server --features tee_quote
```
yuroitaki, line 34: few questions:

1. Don't we need to kill the previous notary service that is running in the vm?
2. Since we don't have ACME yet, how is this notary server handling TLS?

maceip: this is just a build script -- not to run it in prod, and is designed to support multiple concurrent builds
the tls issue will be addressed in the other pr and will be passed in as the config file so this build script should be oblivious

maceip:
addressed the comments in the pr but merged the ci script into #588  to get that branch e2e working

PR #582, opened 2024-09-11
feat: Add networked interactive verifier example

heeckhau:


heeckhau:
> I'm against including this, we should be looking at how we can reduce code
> 
> Our examples should not be attempting to teach people Rust, how to use server frameworks or how to configure networking. They should focus on demonstrating the key concepts of _our_ API

OK, but we need this to demo `tlsn-js` and we'll need something similar to this for the extension demos too soon. If we don't add here? Where do we add it?

PR #584, opened 2024-09-11
refactor(core): integrate rewrite changes

sinui0:
This PR integrates the changes from #574 into the rest of our crates.


crates/tls/mpc/src/follower.rs
```diff
@@ -60,8 +59,6 @@ pub struct MpcTlsFollower {
 /// Data collected by the MPC-TLS follower.
 #[derive(Debug)]
 pub struct MpcTlsFollowerData {
-    /// The prover's commitment to the handshake data
-    pub handshake_commitment: Option<Hash>,
```
th4s, line 64: What happened to the handshake commitment?

crates/verifier/src/notarize.rs
```diff
@@ -0,0 +1,78 @@
...
+            mut mux_fut,
+            mut vm,
+            mut ot_send,
+            mut ctx,
+            encoder_seed,
+            server_ephemeral_key,
+            connection_info,
+        } = self.state;
+
+        let session_header = mux_fut
```
th4s, line 35: let's rename this to `attestation`.

th4s (approved):
:rocket: very nice :rocket:

some non-blocking questions and comments

crates/core/src/presentation.rs
```diff
@@ -10,7 +12,7 @@ use crate::{
 };
 
 /// A verifiable presentation.
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct Presentation {
     attestation: AttestationProof,
     identity: Option<ServerIdentityProof>,
```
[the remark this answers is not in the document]
th4s: Do you have an example for a case where this makes sense? If there is no server identity proof, I can just do TLSNotary with my own server and prove whatever I want?

crates/tls/mpc/src/follower.rs
th4s: Ok, but where is the equivalent thing? I looked for it but did not find it.

crates/core/src/presentation.rs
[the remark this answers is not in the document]
th4s: :grin: ok fine. Just wanted to make sure I had the right understanding :+1:

PR #588, opened 2024-09-12
feat: tee quote endpoint

maceip:
sgx only, default report data


crates/notary/server/Cargo.toml
```diff
@@ -44,3 +47,4 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 uuid = { workspace = true, features = ["v4", "fast-rng"] }
 ws_stream_tungstenite = { workspace = true, features = ["tokio_io"] }
+sgx-dcap-ql-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.21", optional = true }
```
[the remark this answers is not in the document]
yuroitaki:
```suggestion

sgx-dcap-ql-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.21", optional = true }
```

crates/notary/server/src/server.rs
```diff
@@ -153,7 +156,16 @@ pub async fn run_server(config: &NotaryServerProperties) -> Result<(), NotarySer
         >(notary_globals.clone()))
         .route("/notarize", get(upgrade_protocol))
         .layer(CorsLayer::permissive())
-        .with_state(notary_globals);
+        .with_state(notary_globals.clone());
+
+    #[cfg(feature = "tee_quote")]
```
yuroitaki, line 161: This should be a simpler way, since `quote` needs to use `AuthorizationMiddleware` and doesn't need `NotaryGlobals`

```rust
// line 132
    .route(
            "/info",
            get(|| async move {
                (
                    StatusCode::OK,
                    Json(InfoResponse {
                        version,
                        public_key,
                        git_commit_hash,
                        git_commit_timestamp,
                    }),
                )
                    .into_response()
            }),
    );

    #[cfg(feature = "tee_quote")]
    {
        router = router
            .route("/quote", get(quote));
    }

    let router = router
        .route("/session", post(initialize))
        // Not applying auth middleware to ...
```

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+    response::{IntoResponse, Response},
+};
+use axum_macros::debug_handler;
+use sgx_dcap_ql_rs::{quote3_error_t, sgx_report_t, sgx_target_info_t};
+use tracing::error;
+use tracing::{debug, instrument};
+
+#[instrument(level = "debug", skip_all)]
+pub async fn sgx_quote(
+    State(_notary_globals): State<NotaryGlobals>,
```
yuroitaki, line 15: Can remove `NotaryGlobals` in this file since its not needed

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+    http::StatusCode,
+    response::{IntoResponse, Response},
+};
+use axum_macros::debug_handler;
+use sgx_dcap_ql_rs::{quote3_error_t, sgx_report_t, sgx_target_info_t};
+use tracing::error;
+use tracing::{debug, instrument};
+
+#[instrument(level = "debug", skip_all)]
+pub async fn sgx_quote(
```
yuroitaki, line 14: this function can be private

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+            debug!("Failed to retrieve quote.");
+            Err(quote3_error_t::SGX_QL_ERROR_UNEXPECTED)
+        }
+    }
+}
+
+#[debug_handler(state = NotaryGlobals)]
+pub async fn quote(State(notary_globals): State<NotaryGlobals>) -> Response {
+    let mut target_info: sgx_target_info_t = Default::default();
+    let _result = sgx_dcap_ql_rs::sgx_qe_get_target_info(&mut target_info);
```
yuroitaki, line 41: what does this do when we don't use its result?

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    match sgx_quote(State(notary_globals)).await {
+        Ok(_) => (StatusCode::OK, "Quote retrieved successfully").into_response(),
```
yuroitaki, line 44: don't we need to return the quote to the client?

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+        Err(_) => {
+            NotaryServerError::BadProverRequest("Failed to get quote".to_owned()).into_response()
```
yuroitaki, line 46: it shouldn't be `BadProverRequest` I think since it's not due to client's input, perhaps `Unexpected`? Or do you think there is a need to have a new Error enum called `Tee`?

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+#[instrument(level = "debug", skip_all)]
+pub async fn sgx_quote(
+    State(_notary_globals): State<NotaryGlobals>,
+) -> Result<Option<Vec<u8>>, quote3_error_t> {
+    let sgx_report: sgx_report_t = Default::default();
+
+    let (result, sgx_quote) = sgx_dcap_ql_rs::sgx_qe_get_quote(&sgx_report);
+
+    if result != sgx_dcap_ql_rs::quote3_error_t::SGX_QL_SUCCESS {
+        error!("Failed to retrieve quote");
```
yuroitaki, line 22: should we log the error message here from the sgx library on why it fails?

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+        return Err(quote3_error_t::SGX_QL_ERROR_UNEXPECTED);
+    }
+
+    match sgx_quote {
+        Some(q) => {
+            debug!("Quote data: {:?}", q);
+            Ok(Some(q))
```
yuroitaki, line 29: we can just directly return `q` without wrapping it in `Option` right?

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,49 @@
...
+use tracing::error;
+use tracing::{debug, instrument};
+
+#[instrument(level = "debug", skip_all)]
+pub async fn sgx_quote(
+    State(_notary_globals): State<NotaryGlobals>,
+) -> Result<Option<Vec<u8>>, quote3_error_t> {
+    let sgx_report: sgx_report_t = Default::default();
+
+    let (result, sgx_quote) = sgx_dcap_ql_rs::sgx_qe_get_quote(&sgx_report);
```
yuroitaki, line 19: question: does this sgx library have async ability?

yuroitaki (changes requested):
Nice stuff, left some comments :)

yuroitaki (changes requested):
<ignore this, vs code github plugin bug>

crates/notary/server/Cargo.toml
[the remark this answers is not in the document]
maceip: > nice find! can we leave a comment here on why we choose this repo instead of the other listed in https://github.com/intel/linux-sgx?tab=readme-ov-file#introduction? just for future reference :)

Which other libs? The other things mentioned at your link are a kernel module for older versions of Linux,  a patched openssl impl, a k8s gadget, etc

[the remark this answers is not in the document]
yuroitaki: > > nice find! can we leave a comment here on why we choose this repo instead of the other listed in https://github.com/intel/linux-sgx?tab=readme-ov-file#introduction? just for future reference :)
> 
> Which other libs? The other things mentioned at your link are a kernel module for older versions of Linux, a patched openssl impl, a k8s gadget, etc

no referring to any specific one, but just thought a comment in the `Cargo.toml` explaining why we picked this `SGXDataCenterAttestationPrimitives` repo will be informative

crates/notary/server/src/tee.rs
```diff
@@ -1,49 +1,53 @@
...
+    //// https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf
     let _result = sgx_dcap_ql_rs::sgx_qe_get_target_info(&mut target_info);
 
     match sgx_quote(State(notary_globals)).await {
-        Ok(_) => (StatusCode::OK, "Quote retrieved successfully").into_response(),
-        Err(_) => {
-            NotaryServerError::BadProverRequest("Failed to get quote".to_owned()).into_response()
-        }
+        Ok(quote) => (StatusCode::OK, Json(QuoteBytesResponse { quote })).into_response(),
+        Err(code) => (StatusCode::INTERNAL_SERVER_ERROR, (code as u8).to_string()).into_response(),
```
yuroitaki, line 51: is this error code safe to be returned to client? ie will it contain any sensitive info?

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,64 @@
+name: sgx-build-azure-ssh
+
+on:
+  push:
+    branches: [ "quote-presentation" ]
+  pull_request:
+    branches: [ "quote-presentation" ]
```
heeckhau, line 7: I assume this temporary?

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,64 @@
...
+            notarypid=$!
+            win=0
+            SECONDS=0; set -o pipefail; while [ $SECONDS -lt 300 ]; do echo -e "GET /info HTTP/1.1\r\nHost: localhost\r\nConnection: Close\r\n\r\n" | openssl s_client -quiet localhost:$random_port 2>&1 | tee s.log; if [ $? -eq 0 ]; then win=1; break; fi; sleep 2; done
+            if [ $win -eq 0 ]; then echo "Build Failed"; kill -SIGTERM $notarypid; exit 1; fi
+            cat s.log | grep '^{' |  jq '. | tostring' > quotejson
+            mapfile quote < quotejson
+            curl -L -X PATCH -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $PAT_TOKEN" -H "Connection: close" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/$GITHUB_REPOSITORY/environments/TEE/variables/AZURE_TEE_BUILD_PORT -d "{\"name\":\"AZURE_TEE_BUILD_PORT\",\"value\":$quote}"
+            kill -SIGTERM $notarypid
+            sleep 5
+            exit 0
```
heeckhau, line 53: I'd prefer this to be a real script (e.g. in `cd-scripts`), so the same code can be used for development and ci.

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,64 @@
...
+            sed -i '/port*: *7047/s/7047/'$random_port'/' ../config.yaml
+            make clean
+            SGX=1 make start-gramine-server &
+            notarypid=$!
+            win=0
+            SECONDS=0; set -o pipefail; while [ $SECONDS -lt 300 ]; do echo -e "GET /info HTTP/1.1\r\nHost: localhost\r\nConnection: Close\r\n\r\n" | openssl s_client -quiet localhost:$random_port 2>&1 | tee s.log; if [ $? -eq 0 ]; then win=1; break; fi; sleep 2; done
+            if [ $win -eq 0 ]; then echo "Build Failed"; kill -SIGTERM $notarypid; exit 1; fi
+            cat s.log | grep '^{' |  jq '. | tostring' > quotejson
+            mapfile quote < quotejson
+            curl -L -X PATCH -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $PAT_TOKEN" -H "Connection: close" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/$GITHUB_REPOSITORY/environments/TEE/variables/AZURE_TEE_BUILD_PORT -d "{\"name\":\"AZURE_TEE_BUILD_PORT\",\"value\":$quote}"
```
heeckhau, line 50: This is confusing. Is this what will be replaced with `AZURE_TEE_BUILD_ATTESTATION`?

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,64 @@
...
+            kill -SIGTERM $notarypid
+            sleep 5
+            exit 0
+    - name: ✨ fet quotech from gh, write it to runner
```
heeckhau, line 54: "Fetch quote"?

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,64 @@
...
+      run:  |
+        curl -L -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${{ secrets.PAT_SET_ENV }}" -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/$GITHUB_REPOSITORY/environments/TEE/variables/AZURE_TEE_BUILD_PORT > /home/runner/work/_temp/sgx-build-quote.txt
```
heeckhau, line 56: Why `curl` here? This is running on GH, so I assume you have direct access to `AZURE_TEE_BUILD_PORT`

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - name: upload it
+      uses: actions/upload-artifact@v4
+      with:
+        path: /home/runner/work/_temp/sgx-build-quote.txt
+    - name: get github to sign our measurement
```
heeckhau, line 61: Is the order important here? Shouldn't you sign before you upload?

heeckhau (changes requested):

maceip:
[sgx-build-azure-ssh](https://github.com/tlsnotary/tlsn/actions/workflows/ssh-azure-tee-build.yaml)

crates/notary/server/openapi.yaml
```diff
@@ -206,6 +206,9 @@ components:
         gitCommitTimestamp:
           description: The git commit timestamp of source code that this notary server is running
           type: string
+        quote:
+            description: a measurement attesting to the hardware, if available
```
yuroitaki, line 210:
```suggestion
            description: A measurement attesting to the hardware, if available
```

yuroitaki: would `teeQuote` be clearer as a name instead of just `quote`?

yuroitaki: also we need to define the fields of this quote object in this yaml file right

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,149 @@
...
+    //// Writing 64 zero bytes to the gramine report pseudo-hardware `/dev/attestation/user_report_data`
+    let mut report_data_file = OpenOptions::new()
+        .write(true)
+        .open("/dev/attestation/user_report_data")?;
+    report_data_file.write_all(&[0u8; 64])?;
+
+    //// Reading from the gramine quote pseudo-hardware `/dev/attestation/quote`
+    let mut quote_file = File::open("/dev/attestation/quote")?;
+    let mut quote = Vec::new();
+    quote_file.read_to_end(&mut quote)?;
```
yuroitaki, line 88: question: is the quote measurement already available when this read happens? or this read ll trigger the measurement to happen?

crates/notary/server/src/domain.rs
```diff
@@ -16,4 +17,7 @@ pub struct InfoResponse {
     pub git_commit_hash: String,
     /// Current git commit timestamp of notary-server
     pub git_commit_timestamp: String,
+    /// hardware attestation
```
yuroitaki, line 20:
```suggestion
    /// Hardware attestation
```

crates/notary/server/src/server.rs
```diff
@@ -46,10 +46,18 @@ use crate::{
     util::parse_csv_file,
 };
 
+#[cfg(feature = "tee_quote")]
+use crate::tee::{ephemeral_keypair, quote};
+
 /// Start a TCP server (with or without TLS) to accept notarization request for both TCP and WebSocket clients
 #[tracing::instrument(skip(config))]
 pub async fn run_server(config: &NotaryServerProperties) -> Result<(), NotaryServerError> {
+    //tee uses ephemeral key
```
yuroitaki, line 55:
```suggestion
    // tee uses ephemeral key
```

yuroitaki (changes requested):
Did one pass on the rust bit; will review the other bits later :) very nice stuffs!

crates/notary/server/config/gramine/Makefile
```diff
@@ -0,0 +1,62 @@
+# notary-server testing only
```
yuroitaki, line 1: is it possible to put this gramine folder in a `tee` parent folder instead of `config`? just to be more distinctive

.github/workflows/ssh-azure-tee-build.yaml
yuroitaki: > "Fetch quote"?

is this right?

crates/notary/server/config/gramine/target
```diff
@@ -0,0 +1 @@
+../../../../../target
```
yuroitaki, line 1: we need this file?

.github/workflows/ssh-azure-tee-build.yaml
```diff
@@ -0,0 +1,88 @@
...
+    environment: tee
+    steps:
+    - name: azure-tee-build-${{ github.sha }}
+      uses: appleboy/ssh-action@v1.0.3
+      env:
+         # using personal access token until devops links acct
+         PAT_TOKEN: ${{ secrets.PAT_SET_ENV }}
+      with:
+         host: ${{ secrets.AZURE_TEE_BUILD_HOST }}
+         username: ${{ secrets.AZURE_BUILD_TEE_USERNAME }}
```
yuroitaki, line 41:
```suggestion
         username: ${{ secrets.AZURE_TEE_BUILD_USERNAME }}
```

yuroitaki (changes requested):
🔥

.github/workflows/ssh-azure-tee-build.yaml
yuroitaki: > Why `curl` here? This is running on GH, so I assume you have direct access to `AZURE_TEE_BUILD_PORT`

same question here

crates/notary/server/src/domain.rs
```diff
@@ -16,4 +17,7 @@ pub struct InfoResponse {
     pub git_commit_hash: String,
     /// Current git commit timestamp of notary-server
     pub git_commit_timestamp: String,
+    /// Hardware attestation
```
yuroitaki, line 20: what happened to the changes in `openapi.yaml`? the response should still be updated

crates/notary/server/fixture/tee/Dockerfile
```diff
@@ -0,0 +1,25 @@
+#tlsnotary server for testing <> gramine sgx (gramine1.7, g++13, libiomp off :()
```
yuroitaki, line 1: any reason ure putting these tee files in `fixture`? asking this cause `fixture` here is mainly for mock data for testing or local trial purposes

yuroitaki: `crates/notary/server/tee`

crates/notary/server/src/tee.rs
```diff
@@ -0,0 +1,149 @@
...
+    path::Path,
+};
+use tracing::{debug, error, instrument};
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Quote {
+    raw_quote: Option<String>,
+    mrsigner: Option<String>,
+    mrenclave: Option<String>,
+    error: Option<String>,
```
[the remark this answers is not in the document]
yuroitaki: Discussed and decided that there might be clients who dont care about TEE errors and so we shouldnt return 500 if only TEE fails in this `/info` endpoint

PR #592, opened 2024-09-19
Remove clippy::blocks_in_conditions

valo:
The bug in clippy is fixed in rust 1.81, so these are not needed any more


yuroitaki (approved):
@valo thanks for the PR, looks good to me!

PR #600, opened 2024-09-23
feat(wasm): clone getters

### LINKED ISSUE #598: bug: cannot serialize Attestation and Secrets from NotarizationOutputs
https://github.com/tlsnotary/tlsn/blob/dev/crates/wasm/src/prover/mod.rs#L138-L141

For some reason a json object is returned here instead of a serializable class

Screenshot 2024-09-20 at 4 15 24 PM

sinui0:
Closes #598 

This adds `Clone` derives on some fields and modifies the `NotarizationOutput` exported type to JS to be a class.


heeckhau (approved):
🙏

PR #607, opened 2024-09-26
Coverage reports in CI

heeckhau:


.github/workflows/ci.yml
```diff
@@ -119,3 +119,23 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install stable rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Generate code coverage
+        run: cargo llvm-cov --all-features --workspace --exclude tlsn-tls-client --exclude tlsn-tls-core --html
```
yuroitaki, line 135: --all-targets < do we need this?

yuroitaki (commented):
Nice! Do we need to run this llvm-cov in release mode so that its faster?

yuroitaki (approved):
Nice!

PR #610, opened 2024-09-30
fix(core): remove serde flatten

### LINKED ISSUE #609: bug(core): Presentation is not serializable
> Mm, this is rust move semantics not translating well into JS. The presentation instance is actually consumed when you call verify on it. I'll change that, but it introduces more cloning.

btw if i call `serialize()` first, i get the `SequenceMustHaveLength` errors below

Screenshot 2024-09-30 at 5 18 06 PM

_Originally posted by @0xtsukino in https://github.com/tlsnotary/tlsn/issues/600#issuecomment-2382587762_

sinui0:
Closes #609

### DIFF
--- a/crates/core/src/merkle.rs
+++ b/crates/core/src/merkle.rs
@@ -20,7 +20,6 @@ impl MerkleError {
 pub(crate) struct MerkleProof {
     alg: HashAlgId,
     tree_len: usize,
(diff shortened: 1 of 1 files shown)


yuroitaki (approved):
👍

heeckhau (approved):

PR #613, opened 2024-09-30
docs(core): additional documentation and examples

sinui0:
This PR adds more documentation to the core crate, including some examples. To simplify the example for verifying a presentation I added a method for fetching the key.


crates/core/src/attestation/proof.rs
```diff
@@ -98,6 +103,10 @@ impl BodyProof {
         Ok(BodyProof { body, proof })
     }
```
yuroitaki, line 105: Need documentation comment for this func?

yuroitaki (commented):
Nice! just one small comment.

crates/core/src/presentation.rs
```diff
@@ -1,4 +1,26 @@
...
+//! > that is synthesized from, but do not contain, the original verifiable
+//! > credentials (for example, zero-knowledge proofs).
+//!
+//! Instead of a credential, a presentation in this context is a proof of an
+//! attestation from a Notary along with additional selectively disclosed
+//! information about the TLS connection such as the server's identity and the
+//! application data communicated with the server.
+//!
+//! A presentation is self-contained and can be verified by a Verifier without
+//! needing access to external data. The Verifier need only check that the key
```
th4s, line 20: Is this colloquial? Shouldn't this be
```
The Verifier only needs to check ...
```

th4s (approved):
Nice doc :rocket:

crates/core/src/attestation/proof.rs
yuroitaki: oh weird, vscode plugin acting weird, i'm referring to L126

Issue #612, opened 2024-09-30
feat(notary-client): Add support for specifying path


sinui0:
The Notary client needs to support specifying the path when connecting to a Notary. For example, our reverse proxy routes to different instances based on the version: `https://notary.pse.dev/v0.1.0-alpha.6`. Right now there is no way to use that.

yuroitaki:
Fixed as per https://github.com/tlsnotary/tlsn/pull/614

PR #614, opened 2024-09-30
Fixed client issue of not being able to specify the notary url path

ElusAegis:
Fix for the issue #612 

Add an optional configuration `path` to the `NotaryClient` structure, to supliment the existing `port` and `host`.  Defaults to empty String.


notary/client/src/client.rs
```diff
@@ -116,6 +116,9 @@ pub struct NotaryClient {
     /// Port of the notary server endpoint.
     #[builder(default = "self.default_port()")]
     port: u16,
+    /// Path of the notary server endpoint.
```
yuroitaki, line 119:
```suggestion
    /// Path of the notary server endpoint, e.g. version.
```

yuroitaki (changes requested):
LGTM except for one minor thing — thanks for the PR!

notary/client/src/client.rs
ElusAegis: Thank you, adjusted it

crates/notary/client/src/client.rs
```diff
@@ -113,6 +113,9 @@ pub struct NotaryClient {
     /// Port of the notary server endpoint.
     #[builder(default = "self.default_port()")]
     port: u16,
+    /// URL path of the notary server endpoint, e.g. version.
```
yuroitaki, line 116:
```suggestion
    /// URL path prefix of the notary server endpoint, e.g. "https://<host>:<port>/<path_prefix>/...".
```

crates/notary/client/src/client.rs
```diff
@@ -113,6 +113,9 @@ pub struct NotaryClient {
...
+    #[builder(setter(into), default = "String::from(\"\")")]
+    path: String,
```
yuroitaki, line 118: let's name this `path_prefix` to be clearer

crates/notary/client/src/client.rs
```diff
@@ -201,6 +204,11 @@ impl NotaryClient {
         notarization_request: NotarizationRequest,
     ) -> Result<(S, String), ClientError> {
         let http_scheme = if self.tls { "https" } else { "http" };
+        let path = if self.path.is_empty() || self.path.starts_with('/') {
```
yuroitaki, line 207: nit: good idea with the 2nd condition, but i prefer to be stricter with our api design, to have simpler logic and avoid confusion with too many patterns (esp. if the upside isn't significant) — i'd only keep the first condition

yuroitaki (changes requested):

crates/notary/client/src/client.rs
ElusAegis: Makes sense , that is a lot more undestandable

crates/notary/client/src/client.rs
ElusAegis: Agree, even if `/` does already exist, most servers will change resulting `//` into `/` - I have checked and PSE server already does

yuroitaki (approved):
LGTM! thanks again @ElusAegis

yuroitaki:
@ElusAegis the formatting ci failed, can you follow the instructions here? https://github.com/tlsnotary/tlsn/blob/dev/CONTRIBUTING.md#pull-requests

ElusAegis:
@yuroitaki please have a look - I have fixed the formatting

yuroitaki:
> @yuroitaki please have a look - I have fixed the formatting

Merged! Thanks for your help~

PR #619, opened 2024-10-02
Add convenience script to run most ci checks locally

heeckhau:
This script can be used to run checks before pushing changes to the repository.
It is a good approximation of what CI will do.

### DIFF
--- a/pre-commit-check.sh
+++ b/pre-commit-check.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+# This script is used to run checks before committing changes to the repository.
+# It is a good approximation of what CI will do.
+
+# Fail if any command fails
+set -e
+
+# Check formatting
+cargo +nightly fmt --all
+
+# Check clippy
+cargo clippy --all-features --all-targets -- -D warnings
+
+# Build all targets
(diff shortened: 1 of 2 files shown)


pre-commit-check.sh
```diff
@@ -0,0 +1,22 @@
...
+cargo clippy --all-features --all-targets -- -D warnings
+
+# Build all targets
+cargo build --all-targets
+
+# Run tests
+cargo test
+
+# Run wasm tests
+./crates/wasm-test-runner/run.sh
```
yuroitaki, line 22: not running those ignored expensive integration tests?

heeckhau: I was a bit undecided here. I wanted a script to "quickly check" my local branch before pushing to GitHub and avoiding follow-up commits.

The main reason for extra follow-up commits is mostly **formatting** and **clippy**. So maybe we should limit this script to those.

yuroitaki: agreed, we can let testing be done in GH and only use this for clippy and formatting.

themighty1 (approved):
Thanks, that's a good start. The name of the script is appropriate, too.

PR #629, opened 2024-10-08
ci: use rust containers & sccache 

maceip:
demo micro-perf gains; ymmv =>

- use official rust containers (& mirrored nightlies) 
- use mozilla's sccache with gha backend


.github/workflows/ci.yml
```diff
@@ -67,33 +65,28 @@ jobs:
...
-        with:
-          targets: wasm32-unknown-unknown
-          toolchain: stable
-
-      - name: Install nightly rust toolchain
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: wasm32-unknown-unknown,x86_64-unknown-linux-gnu
-          toolchain: nightly
-          components: rust-src
```
heeckhau, line 85: I assume `rust-src` was added for a reason. Let's see if the build breaks:
```
rustup component add rust-src --toolchain nightly --target wasm32-unknown-unknown
rustup component add rust-src --toolchain nightly --target x86_64-unknown-linux-gnu
```

.github/workflows/ci.yml
```diff
@@ -67,68 +63,64 @@ jobs:
...
-        uses: Swatinem/rust-cache@v2.7.3
+        run: cargo install wasm-pack
 
       - name: Run tests
         run: |
-          cd crates/wasm-test-runner
-          ./run.sh
+          
+          cd crates/wasm-test-runner 
+          chmod +x run.sh && ./run.sh
```
heeckhau, line 93: Is `chmod +x run.sh` necessary here? It should already be executable

heeckhau (approved):

codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
```
heeckhau, line 1: @maceip I added this coverage config file here. Because the extra ".no_sandbox()" made code coverage checks fail 🤷

.github/workflows/ci.yml
```diff
@@ -19,45 +19,41 @@ env:
   #   - https://github.com/privacy-scaling-explorations/mpz/issues/178
   # 32 seems to be big enough for the foreseeable future
   RAYON_NUM_THREADS: 32
+  # these remove the need for git in notary-server build.rs:
+  GIT_COMMIT_HASH: ${{ github.event.pull_request.head.sha || github.sha }}
+  GIT_COMMIT_TIMESTAMP: ${{ github.event.repository.updated_at}}
```
yuroitaki, line 24: does `github.event.repository.updated_at` always refer to the timestamp of the head commit? It seems like the timestamp ll also be updated if someone updates e.g. the description of the repo (https://github.com/orgs/community/discussions/24442)

yuroitaki: i coulnd't find the documentation for `github.event.repository.updated_at`, the link above is the closest i can find — worth double checking imo

heeckhau:
The builds are slower

image

image

.github/workflows/ci.yml
heeckhau, line 24: Good catch
```suggestion
  GIT_COMMIT_TIMESTAMP: ${{ github.event.repository. pushed_at}}
```

yuroitaki: @heeckhau seems like `repo.pushed_at` is for commit on *any branch* — `pushed_at will be updated any time a commit is pushed to any of the repository’s branches.`, which is also wrong?

PR #643, opened 2024-10-16
Do not include git commit timestamp in notary server + use git2 instead of git command + add dirty suffix

heeckhau:
There are some problems with the git commit timestamp (GitHub actions has no exact variable for this, it might be different across time zones,...) so I decided to drop this field.

I also used Rust's git2 instead of calling out to a potentially missing git binary.

For local builds, I also added a flag to indicated the build might have local changes (wrt to the commit hash)


crates/notary/server/src/domain.rs
```diff
@@ -14,6 +14,4 @@ pub struct InfoResponse {
     pub public_key: String,
     /// Current git commit hash of notary-server
     pub git_commit_hash: String,
-    /// Current git commit timestamp of notary-server
```
yuroitaki, line 17: need to update openapi.yaml too for this api response

yuroitaki (changes requested):
Great stuff! Though when i built and ran the notary server as docker container locally, the '/info' endpoint returns `(with local changes)` eventhough i never modified anything. Does that happen at your end too?

heeckhau:
> Great stuff! Though when i built and ran the notary server as docker container locally, the '/info' endpoint returns `(with local changes)` eventhough i never modified anything. Does that happen at your end too?

It should only add the suffix if there are actual changes. I struggled more than expect on the implementation. If it does not work for you, we may better drop it

maceip:
lgtm => in rust:slim:

```
# git
sh: 43: git: not found
# cargo run --bin notary-server
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/notary-server`
2024-10-18T15:37:18.026878Z DEBUG main ThreadId(01) notary_server: Server config loaded
```

maceip (approved):

yuroitaki (approved):
LGTM!

crates/notary/server/build.rs
```diff
@@ -1,30 +1,49 @@
...
+        match get_commithash_with_dirty_suffix() {
+            Ok(commit_hash_with_suffix) => {
+                // Pass value as env var to the notary server
+                println!("cargo:rustc-env=GIT_COMMIT_HASH={commit_hash_with_suffix}");
+            }
+            Err(e) => {
+                eprintln!("Failed to get commit hash in notary server build");
+                eprintln!("Fix the error or configure GIT_COMMIT_HASH as environment variable");
+                return Err(e.message().into());
+                // return Err(e.to_string().into());
```
yuroitaki, line 16: need this?

PR #642, opened 2024-10-16
chore: Bump version to 0.1.0-alpha.8-pre

heeckhau:

PR #644, opened 2024-10-17
Version bump correction

heeckhau:
#642


th4s (commented):
Thanks :+1:

but this does not address those two I mentioned in #642?

th4s (approved):

PR #645, opened 2024-10-17
Remove unnecessary dependency from notary-client crate

0xSachinK:
Removed `tlsn-common` as a dependency from the `notary-client` crate since it wasn't being used. It was likely a leftover from the v7 refactor.


heeckhau:
Thanks, but this was recently fixed in https://github.com/tlsnotary/tlsn/pull/638

PR #648, opened 2024-10-22
Fix/core validation

themighty1:
This PR adds validation to some of the core types.

Also renames `tree_len` to `leaf_count` since "tree length" is too ambiguous.

### DIFF
--- a/crates/core/src/transcript/encoding/proof.rs
+++ b/crates/core/src/transcript/encoding/proof.rs
@@ -27,6 +27,7 @@ opaque_debug::implement!(Opening);
 
 /// An encoding commitment proof.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(try_from = "validation::EncodingProofUnchecked")]
 pub struct EncodingProof {
     /// The proof of inclusion of the commitment(s) in the Merkle tree of
     /// commitments.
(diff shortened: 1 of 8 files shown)


crates/core/src/transcript/encoding/proof.rs
```diff
@@ -182,3 +182,92 @@ impl From<MerkleError> for EncodingProofError {
...
+#[derive(Debug, thiserror::Error)]
+#[error("invalid encoding proof: {0}")]
+pub struct InvalidEncodingProof(&'static str);
+
+mod validation {
+    use super::*;
+
+    /// The maximum allowed height of the Merkle tree of encoding commitments.
+    ///
+    /// The statistical security parameter (SSP) of the encoding commitment
```
yuroitaki, line 196: just curious on this statement about how SSP is calculated for the encoding commitment protocol: is there a paper/textbook that outlines this or its our team's finding?

crates/core/src/transcript/encoding/proof.rs
```diff
@@ -182,3 +182,92 @@ impl From<MerkleError> for EncodingProofError {
...
+    /// protocol is calculated as "the number of uniformly random bits in a
+    /// single bit's encoding minus `MAX_HEIGHT`".
+    ///
+    /// For example, a bit encoding used in garbled circuits typically has 127
+    /// uniformly random bits, hence when using it in the encoding
+    /// commitment protocol, the SSP is 117 bits.
+    ///
+    /// DO NOT use bit encodings which have less than 50 uniformly random bits,
+    /// since the SSP < 40 bits is widely considered inadequate.
+    const MAX_HEIGHT: usize = 10;
```
yuroitaki, line 206: is there any reason why 10 is chosen?

yuroitaki (approved):
LGTM! just some questions

crates/core/src/transcript/encoding/proof.rs
themighty1: This is not documented. Maybe we should document it. But this commitment protocol is kinda still not solidifed to document it atm.

The SSP of our encoding commitment protocol is essentially "what is the probabily that a prover who tries to guess the encoding of a single bit will succeed". The success probability is 1/2^SSP.

Suppose, our encodings are 40 bits, then the probability of the prover guessing an encoding of a sinlge bit is 1/2^40. That gives us protocol 40 bits of statistical security.

Suppose, a malicious prover fills the Merkle tree with 2^10 1-bit encodings which he guessed at random. Later, when the prover learns what the encoding seed is, he will check if any of his 2^10 guesses matches the encoding and. If it matches, then the attack succeeded and the prover will reveal that specific Merkle tree leaf which matched to the verifier.

The fact the the prover was allowed to make 2^10 guesses brought down the SSP from 40 to 30. Hence, we need to put a cap on the size of the Merkle tree and also require that the bit-length of a single encoding is not too small. We need to achieve at least SSP = 40.

crates/core/src/transcript/encoding/proof.rs
themighty1: No specific reason other than we want to achieve SSP=40 and don't want to mandate that encodings are unnecessarily large. It seemes reasonable that there won't be 1024 commitments. Although, now I'm realizing that with our json commit-to-everything strategy 1024 may be insufficient.

Let's tag @sinui0 to double check.

crates/core/src/transcript/encoding/proof.rs
yuroitaki: tthanks for the explanation!

crates/core/src/transcript/encoding/proof.rs
themighty1: i bumped it to 30 and left it there as a fail-safe

themighty1:
ready for review @sinui0

PR #652, opened 2024-10-24
Return valid json in Server fixture

heeckhau:
* return json instead of string
* removed trailing whitespace
* use a constant for the default port
* give binary a better name


crates/server-fixture/server/Cargo.toml
```diff
@@ -14,9 +14,14 @@ hyper-util = { workspace = true, features = ["full"] }
...
+serde_json = { workspace = true }
 tlsn-server-fixture-certs = { workspace = true }
 
 [[bin]]
-name = "main"
+name = "tlsn-server-fixture"
 path = "src/main.rs"
+
+[dev-dependencies]
+axum-test = "16.2.0"
```
yuroitaki, line 25: nit: i think we're adopting the format of always using `{ version = "16.2.0" }` in the project?

crates/server-fixture/server/src/lib.rs
```diff
@@ -94,10 +97,18 @@ async fn bytes(
...
+        eprintln!("Failed to parse JSON data: {}", e);
+        StatusCode::INTERNAL_SERVER_ERROR
+    })?))
+}
+
 async fn json(
     State(state): State<Arc<Mutex<AppState>>>,
     Query(params): Query<HashMap<String, String>>,
-) -> Result<Json<&'static str>, StatusCode> {
+) -> Result<Json<Value>, StatusCode> {
```
yuroitaki, line 111: Does any downstream client using this server fixture need to change their handling of the response?

crates/server-fixture/server/src/lib.rs
```diff
@@ -125,3 +136,58 @@ async fn html(
...
+        let state = AppState {
+            shutdown: Some(sender),
+        };
+        app(state)
+    }
+
+    #[tokio::test]
+    async fn hello_world() {
+        let response = get_app()
+            .oneshot(Request::builder().uri("/").body(Body::empty()).unwrap())
```
yuroitaki, line 161: this definitely worked for testing purely the request and response; tho if we want more of an integration test of the server, then the test should spin up the server like its `main.rs` (using `tokio::spawn` & `TcpListener`)

yuroitaki (commented):
nice stuff!

crates/server-fixture/server/Cargo.toml
heeckhau: fixed

crates/server-fixture/server/src/lib.rs
heeckhau: All tests are green. So should be fine

PR #653, opened 2024-10-24
feat: compress partial transcript

themighty1:
This PR enables the prover to compress the partial transcript when sending it to the verifier.

Also renames the field `PartialTranscript::sent_authed` to `sent_authed_idx` to minimize cognitive burden when analyzing the compression logic.


yuroitaki (approved):
Great work!

PR #656, opened 2024-10-24
Use a local server (fixture) for the attestation example

heeckhau:
Use server fixture for tlsn examples + removed Discord example

The attestation example now has different modes: json, html and authenticated


th4s:
I tried to follow the instructions but got
```
Error: Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }
when executing `attestation_prove`.

**Comment by th4s** [comment]
Now it works ^^, forgot `PORT` :)

File: crates/examples/attestation/README.md [hunk 1]
```diff
@@ -1,75 +1,109 @@
...
+
+The examples above demonstrate TLSNotary with public data. TLSNotary can also be used for private data by adding the necessary headers (e.g., an authentication token) or cookies to the request. To run a private data example:
+```shell
+# notarize
+SERVER_PORT=4000 cargo run --release --example attestation_prove -- authenticated
+# present
+cargo run --release --example attestation_present -- authenticated
+# verify
+cargo run --release --example attestation_verify -- authenticated
+```
```
th4s, line 109: In the first example there are already redacted passages, which I would consider private data? My understanding is that you want to emphasize that the redaction feature can be used to hide authentication secrets like passwords, cookies. In this case, I would make it more clear here.

crates/examples/attestation/present.rs
```diff
@@ -26,14 +47,45 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     builder.reveal_sent(&request.request.target)?;
     // Reveal all headers except the value of the User-Agent header.
```
th4s, line 48: This comment needs to be adapted.

crates/examples/attestation/present.rs
```diff
@@ -26,14 +47,45 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let response = &transcript.responses[0];
+    builder.reveal_recv(&response.without_data())?;
+    for header in &response.headers {
+        builder.reveal_recv(header)?;
+    }
+
+    let content = &response.body.as_ref().unwrap().content;
+    match content {
+        tlsn_formats::http::BodyContent::Json(json) => {
+            let reveal_all = false;
```
th4s, line 75: I assume that `reveal_all = false` is for experimentation?  Maybe add a comment to make this clear?

crates/examples/interactive/interactive.rs
```diff
@@ -162,23 +176,23 @@ async fn verifier<T: AsyncWrite + AsyncRead + Send + Sync + Unpin + 'static>(
...
 
 /// Redacts and reveals received data to the verifier.
-fn redact_and_reveal_received_data(prover: &mut Prover<Prove>) -> Idx {
+fn redact_ranges_received(prover: &mut Prover<Prove>) -> Idx {
     let recv_transcript = prover.transcript().received();
     let recv_transcript_len = recv_transcript.len();
 
     // Get the received data as a string.
     let received_string = String::from_utf8(recv_transcript.to_vec()).unwrap();
     // Find the substring "illustrative".
```
th4s, line 185: Comment needs to be adapted.

th4s (approved):
Nice!

crates/examples/attestation/README.md
heeckhau: The difference here is that you need an authentication token to access the data. I'll try to explain it better

crates/examples/attestation/present.rs
```diff
@@ -24,16 +45,48 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     builder.reveal_sent(&request.without_data())?;
     // Reveal the request target.
     builder.reveal_sent(&request.request.target)?;
-    // Reveal all headers except the value of the User-Agent header.
+    // Reveal all headers except the values of the User-Agent and Authorization.
```
heeckhau, line 48:
```suggestion
    // Reveal all headers except the values of User-Agent and Authorization.
```

Issue #655, opened 2024-10-24
make reveal API more flexible


themighty1:
Currently if the prover has commitments to two or more rangesets and they want to reveal them, they must make two or more separate calls to `TranscriptProofBuilder::reveal`.
We want to allow the prover to make a single call to `reveal` passing in the union of all rangesets to be revealed.

Reveal logic is located here: https://github.com/tlsnotary/tlsn/blob/1d663596c1bff07ce307ffd4aa748f083b379eae/crates/core/src/transcript/proof.rs#L215

PR #664, opened 2024-11-29
Add reveal groups of ranges.

yuroitaki:
Addresses https://github.com/tlsnotary/tlsn/issues/655

### DIFF
--- a/crates/core/src/transcript/proof.rs
+++ b/crates/core/src/transcript/proof.rs
@@ -281,6 +281,22 @@ impl<'a> TranscriptProofBuilder<'a> {
         self.reveal(ranges, Direction::Sent)
     }
 
+    /// Reveals multiple groups of ranges given in the sent transcript using the
+    /// default kind of commitment.
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges_groups` - Groups of ranges to reveal.
+    pub fn reveal_sent_multi(
+        &mut self,
+        ranges_groups: &[&dyn ToRangeSet<usize>],
+    ) -> Result<&mut Self, TranscriptProofBuilderError> {
+        for ranges in ranges_groups.iter() {
+            self.reveal(*ranges, Direction::Sent)?;
+        }
+        Ok(self)
+    }
+
     /// Reveals the given ranges in the received transcript using the default
     /// kind of commitment.
     ///
@@ -292,6 +308,22 @@ impl<'a> TranscriptProofBuilder<'a> {
         ranges: &dyn ToRangeSet<usize>,
     ) -> Result<&mut Self, TranscriptProofBuilderError> {
         self.reveal(ranges, Direction::Received)
+    }
+
+    /// Reveals multiple groups of ranges given in the received transcript using
+    /// the default kind of commitment.
+    ///
+    /// # Arguments
+    ///
+    /// * `ranges_groups` - Groups of ranges to reveal.
+    pub fn reveal_recv_multi(
+        &mut self,
+        ranges_groups: &[&dyn ToRangeSet<usize>],
+    ) -> Result<&mut Self, TranscriptProofBuilderError> {
+        for ranges in ranges_groups.iter() {
+            self.reveal(*ranges, Direction::Received)?;
+        }
+        Ok(self)
     }
 
     /// Builds the transcript proof.
--- a/crates/examples/attestation/present.rs
+++ b/crates/examples/attestation/present.rs
@@ -8,6 +8,7 @@ use tlsn_examples::ExampleType;
 use tlsn_formats::http::HttpTranscript;
 
 use clap::Parser;
+use utils::range::ToRangeSet;
 
 #[derive(Parser, Debug)]
 #[command(version, about, long_about = None)]
@@ -41,10 +42,11 @@ async fn create_presentation(example_type: &ExampleType) -> Result<(), Box<dyn s
     let mut builder = secrets.transcript_proof_builder();
 
     let request = &transcript.requests[0];
-    // Reveal the structure of the request without the headers or body.
-    builder.reveal_sent(&request.without_data())?;
-    // Reveal the request target.
-    builder.reveal_sent(&request.request.target)?;
+
+    // Reveal multiple parts of the request: (1) its structure without the headers
+    // or body, (2) the request target.
+    builder.reveal_sent_multi(&[&request.without_data(), &request.request.target])?;
+
     // Reveal all headers except the values of User-Agent and Authorization.
     for header in &request.headers {
         if !(header
@@ -63,10 +65,16 @@ async fn create_presentation(example_type: &ExampleType) -> Result<(), Box<dyn s
     }
 
     // Reveal only parts of the response
+    // Use a vector to collect the ranges of all these parts before calling
+    // `reveal_recv_multi`
+    let mut recv_ranges: Vec<&dyn ToRangeSet<usize>> = Vec::new();
+
     let response = &transcript.responses[0];
-    builder.reveal_recv(&response.without_data())?;
+    let response_without_data = &response.without_data();
+    recv_ranges.push(response_without_data);
+
     for header in &response.headers {
-        builder.reveal_recv(header)?;
+        recv_ranges.push(header);
     }
 
     let content = &response.body.as_ref().unwrap().content;
@@ -75,18 +83,20 @@ async fn create_presentation(example_type: &ExampleType) -> Result<(), Box<dyn s
             // For experimentation, reveal the entire response or just a selection
             let reveal_all = false;
             if reveal_all {
-                builder.reveal_recv(response)?;
+                recv_ranges.push(response);
             } else {
-                builder.reveal_recv(json.get("id").unwrap())?;
-                builder.reveal_recv(json.get("information.name").unwrap())?;
-                builder.reveal_recv(json.get("meta.version").unwrap())?;
+                recv_ranges.push(json.get("id").unwrap());
+                recv_ranges.push(json.get("information.name").unwrap());
+                recv_ranges.push(json.get("meta.version").unwrap());
             }
         }
         tlsn_formats::http::BodyContent::Unknown(span) => {
-            builder.reveal_recv(span)?;
+            recv_ranges.push(span);
         }
         _ => {}
     }
+
+    builder.reveal_recv_multi(&recv_ranges)?;
 
     let transcript_proof = builder.build()?;
 


yuroitaki:
> This doesn't address the issue, we're not looking for a method to pass in a slice. Instead, we want to address the problem that the user has to specify each range individually which corresponds to a commitment. When the user provides a range set to reveal, we should look into the encoding tree/hash commitments and determine if we can satisfy it with one or more of the available commitments.

Yeap! Got the same feedback from @themighty1 , and am working on it

crates/core/src/transcript/proof.rs
```diff
@@ -208,48 +208,79 @@ impl<'a> TranscriptProofBuilder<'a> {
...
+                    let mut missing_commitment = true;
+                    // Check if there is any committed rangeset in the encoding tree that is a
+                    // subset of the rangeset — if yes, insert them.
+                    for committed_dir_idx in encoding_tree
+                        .transcript_indices()
+                        .into_iter()
+                        .filter(|(dir, _)| *dir == dir_idx.0)
+                    {
+                        if committed_dir_idx.1.is_subset(&dir_idx.1) {
+                            self.encoding_proof_idxs.insert(committed_dir_idx.clone());
```
themighty1, line 225: I didnt review yet, overall the approach is correct, imo, but
we need to make sure that what we are inserting here is not a subset of what is already present in `encoding_proof_idxs`.
Suppose we already inserted 0..100 into `encoding_proof_idxs`. We don't want to insert anything that is a subset of 0..100 since that would unnecessarily increase the proof size and verification time. 

@yuroitaki

yuroitaki: That's a good point — though I believe that problem already exists before this PR? e.g. where user first passes in a committed header range (incl. key and value), and then only the header key range; should this be taken care in a different PR?

themighty1: I would do it in this PR.
Are you saying that there was code in the examples where the user first calls reveal on key+value and then calls reveal on the key? That seems redundant.

themighty1 (changes requested):

themighty1:
I'm sorry @yuroitaki , I left my comment above a few days ago but gh still had it as "pending". Apparently i am not able to leave a single comment without explicitely marking the review as "request changes".

crates/core/src/transcript/proof.rs
yuroitaki: > I would do it in this PR. Are you saying that there was code in the examples where the user first calls `reveal` on `key+value` and then calls `reveal` on the key? That seems redundant.

Ah no code in examples — I was just giving a scenario how this issue theoretically already exists currently (before this PR); because to pass in a subset of what's present in `encoding_proof idxs`, it will mean

1. User first pass in a rangeset, then a subset rangeset (both are/contain committed range(set)) [already possible before this PR]
2. User first pass in a rangeset, then another rangeset that intersects with the former, where the intersection contains committed range(set) [possible after this PR]

I only thought of (1) in my previous comment, and as you said they seemed to be redundant (and unlikely to happen) because it means e.g. user first calls reveal on `key+value` and then calls reveal on the `key`, or user first calls reveal on `header.start..body.end` and then `header.start..header.end`; 

Though now I think about it, (1) might happen more likely after this PR, e.g. first reveal `request.without_data().start..request.without_data().end`, then reveal `request.request.target` — makes sense to look into the check you suggested

themighty1: Thanks for explaining. I see, in the scenario 1) before this PR there would be a secong encoding proof added which is a redundancy that we overlooked.

1) should be permissible behavior. If the user tries to reveal a subset of what already was revealed, it should be a no-op.

yuroitaki: > > I would do it in this PR. Are you saying that there was code in the examples where the user first calls `reveal` on `key+value` and then calls `reveal` on the key? That seems redundant.
> 
> Ah no code in examples — I was just giving a scenario how this issue theoretically already exists currently (before this PR); because to pass in a subset of what's present in `encoding_proof idxs`, it will mean
> 
> 1. User first pass in a rangeset, then a subset rangeset (both are/contain committed range(set)) [already possible before this PR]
> 2. User first pass in a rangeset, then another rangeset that intersects with the former, where the intersection contains committed range(set) [possible after this PR]
> 
> I only thought of (1) in my previous comment, and they seemed to be redundant (and unlikely to happen) because it means e.g. user first calls reveal on `key+value` and then calls reveal on the `key`, or user first calls reveal on `header.start..body.end` and then `header.start..header.end`;
> 
> Though now I think about it, (1) might happen more likely after this PR, e.g. first reveal `request.without_data().start..request.without_data().end`, then reveal `request.request.target` — makes sense to look into the check you suggested

@themighty1 I realised what I said here is mostly wrong and only partially right — when user reveals superset rangesets (union of multiple committed rangesets), (1) and (2) will not insert a subset of what's present in `encoding_proof_idxs`. Instead, the following conditions will:

(1) Among the committed rangesets, there are rangesets that are subset of another; this is common, e.g. `header key+value` and `header key`.
(2) User reveals a superset rangeset (call it `S`), which covers both `header key+value` and `header key`, causing both to be inserted into `encoding_proof_idxs`.

Subsequent revealing of another superset rangeset that is (1) subset of `S` or  (2) intersects `S`, will not insert `header key+value` and `header key` into `encoding_proof_idxs` because both of them are already present in `encoding_proof_idxs`, which is a `HashMap`.

yuroitaki: For that, I have added a subset check now — https://github.com/tlsnotary/tlsn/pull/664/commits/85fb7c67c8acb3d7bb302318f28ac9c3fa562108; though I will need to do the same for hash-based commitments.

crates/core/src/index.rs
```diff
@@ -6,18 +6,19 @@ use crate::{
...
+/// or field id.
 #[derive(Debug, Clone)]
 pub(crate) struct Index<T> {
     items: Vec<T>,
     // Lookup by field id.
     field_ids: HashMap<FieldId, usize>,
-    // Lookup by transcript index.
-    transcript_idxs: HashMap<Idx, usize>,
+    // Lookup by transcript direction and index.
+    transcript_idxs: HashMap<(Direction, Idx), usize>,
```
yuroitaki, line 21: bug: `Direction` needs to be stored alongside `Idx` as the latter doesn't contain any direction info.

crates/core/src/transcript/proof.rs
yuroitaki: > For that, I have added subset pruning for `encoding_proof_idxs` — https://github.com/tlsnotary/tlsn/pull/664/files/aee64d229e3d51d581d018f10db69a7901ccc6b3..85fb7c67c8acb3d7bb302318f28ac9c3fa562108; though I will need to do the same for hash-based commitments.

Done for hash-based commitments!

crates/core/src/index.rs
themighty1: good catch, i'll open an issue lest we forget (but i think this bug was already addressed in one of the open PRs, or maybe it's in my local wip review branch.)

crates/core/src/transcript/proof.rs
```diff
@@ -206,50 +285,77 @@ impl<'a> TranscriptProofBuilder<'a> {
...
+                            missing_commitment = false;
+                        }
+                    }
+                    // If no committed rangeset is a subset, that means the rangeset is missing
+                    // fully or partially in the encoding tree (which means it has not been
+                    // committed).
+                    if missing_commitment {
```
themighty1, line 309: @yuroitaki , a quick observation here (I haven't fully grokked the rest of the logic, so maybe this comment is irrelevant).

Assume there exists a proof in the `encoding_tree` for ranges 0..20. Assume the prover calls reveal(0..25). 
The current logic will insert 0..20 into `encoding_proof_idxs`and will set `missing_commitment` to false.

I think we should make it an error - if the user tries to reveal something which was not committed, we want to error out.

themighty1 (changes requested):

crates/core/src/transcript/proof.rs
yuroitaki: @themighty1 interesting, I was assuming`(0..25)` to be treated as the umbrella/superset range, but if it only reveals a *single* subset range, then it doesn't achieve the purpose of this PR: to support revealing *multiple* subset ranges. I will make the change.

crates/core/src/index.rs
yuroitaki: ah i actually also fixed this bug in this PR itself

crates/core/src/transcript/proof.rs
yuroitaki: Upon clarification, we agreed on making the reveal API to only support revealing the exact rangeset that's passed in. For example, if two rangesets are committed: `[0..2, 8..9], [2..4, 9..11]`, then

(a) revealing `[0..4, 8..11]` is valid
(b) revealing `[0..11]` is not valid

yuroitaki:
Just realised there is one more polishing to do, will notify when it's ready to be reviewed

crates/examples/attestation/present.rs
```diff
@@ -38,13 +39,30 @@ async fn create_presentation(example_type: &ExampleType) -> Result<(), Box<dyn s
...
+    // body. (2) request target.
+    //
+    // R is constructed via the union of ranges of (1) and (2).
+    //
+    // NOTE: Ideally, one should NOT call any `reveal` function more than once
+    // for these superset ranges to avoid unnecessary computation time.
+    //
+    // Instead of `reveal` on 'superset range A', and then `reveal` again
+    // on 'superset range B', one should union all these superset ranges,
+    // and then `reveal` at once.
```
themighty1, line 59: I would avoid putting here the rest of the comment starting from NOTE:
The way that the reveal API works now is the actual expected behaviour.
Talking about "unnecessary computation time" will cause the user to wonder about such small details. But we know that compute overhead is so miniscule, it's not worth drawing attention to.

Additionally, we encourage to use reveal() only once but then immediately a few lines below call it multiple times in this example. This will make the user think the example is wrong/the comment is wrong/the user missed something.

I feel we can allow the user to be able to both call reveal one time or multiple times, depending on their code ergonomics needs.
I feel that the user should be exposed to as few inconsequential details as possible. But feel free to keep the NOTE: if you still think it is valuable.

themighty1 (approved):
Very nicely done!!!

crates/examples/attestation/present.rs
yuroitaki: > I would avoid putting here the rest of the comment starting from NOTE: The way that the reveal API works now is the actual expected behaviour. Talking about "unnecessary computation time" will cause the user to wonder about such small details. But we know that compute overhead is so miniscule, it's not worth drawing attention to.
> 
> Additionally, we encourage to use reveal() only once but then immediately a few lines below call it multiple times in this example. This will make the user think the example is wrong/the comment is wrong/the user missed something.
> 
> I feel we can allow the user to be able to both call reveal one time or multiple times, depending on their code ergonomics needs. I feel that the user should be exposed to as few inconsequential details as possible. But feel free to keep the NOTE: if you still think it is valuable.

Ah i see, i guess the point i was trying to convey wasn't clear: that for superset ranges, they should call `reveal` once after union-ing all of them; but for exact ranges, they can call multiple times. I was mainly worrying about the compute overhead when the following is executed `N` times (the number of times user call `reveal` with superset ranges):

```rust
 for committed_dir_idx in encoding_tree
      .transcript_indices()
          .into_iter()
          .filter(|(dir, _)| *dir == dir_idx.0)
      {
          if committed_dir_idx.1.is_subset(&dir_idx.1) {
              staged_subsets.push(committed_dir_idx);
          }
      }
```

Because if `N` is big, then this block might cause some noticeable overhead, since `is_subset` itself contains an inner loop. But rationally users wouldn't reveal such large `N` — let me try to change the wording to make it clearer and less confusing.

themighty1: thanks, lgtm

themighty1:
lgtm

crates/examples/attestation/present.rs
```diff
@@ -62,17 +63,63 @@ async fn create_presentation(example_type: &ExampleType) -> Result<(), Box<dyn s
...
+        &request
+            .without_data()
+            .union(&request.request.target.to_range_set()),
+    )?;
+
+    /*
+    SIDE NOTE:
+
+    In the code above, we demonstrated how `reveal_sent` can be used on
+    (1) exact range (i.e. each header).
```
heeckhau, line 81:
```suggestion
    (1) exact range (i.e. selected headers).
```

heeckhau (approved):

crates/examples/attestation/present.rs
yuroitaki: it's every header no? just that USER_AGENT and AUTH is without value?

heeckhau: I interpret that as skipping a header :-)
You decide

yuroitaki: good point!

yuroitaki:
@sinui0 nice suggestions! indeed that seems to be simpler and faster — will look into it;

some questions i have meanwhile:

1. On step 2 `check if the provided (Direction, RangeSet) tuple is contained within the...`, by `contained within` you mean via a subset check right?
2. Doesn't step 2 always guarantee that there will always be a solution for the set cover problem at step 4? 
3. On the comment of `this forces them to union the ranges they want to reveal ahead of time then pass to the builder.` — I don't quite get how the current approach has this issue, or how your suggestion differs from the current approach on this aspect, can you elaborate more?

yuroitaki:
@sinui0 i managed to implement the greedy set cover algorithm in https://github.com/tlsnotary/tlsn-utils/pull/54; and integrated it into this PR — let me know what you think :)

Though didn't get to apply it for the hash-based commitments (hence haven't removed my previous PR's implementation), and missing comments still — feel free to take over this if we need to ship this next week when I'm not around.

th4s (approved):

crates/core/src/transcript/proof.rs
```diff
@@ -905,91 +626,8 @@ mod tests {
         if success {
```
yuroitaki, line 627: we should delete this part too since its testing hash commitment, right?

crates/core/src/transcript/proof.rs
```diff
@@ -417,45 +515,120 @@ mod tests {
...
+            assert!(matches!(err.kind, BuilderErrorKind::MissingCommitment));
+        }
 
-        let provider = CryptoProvider::default();
-        let err = transcript_proof
-            .verify_with_provider(&provider, &attestation.body)
-            .err()
-            .unwrap();
-        assert!(matches!(err.kind, ErrorKind::Encoding));
+        // Hash commitment kind
```
yuroitaki, line 602: we should delete this part too since its testing hash commitment, right?

themighty1:
ack

PR #668, opened 2024-12-02
Authdecode

themighty1:
This PR is not yet ready for review.

This PR is a copy of https://github.com/tlsnotary/tlsn/pull/479
I cherry-picked all commits from 479 to create this PR because due to dir structure changes over the last year I had too many headaches with resolving conflicts.

All the feedback left on 478 will be addressed here.

PR668 original description:

This PR implements the AuthDecode protocol to be used to create zk-friendly commitments.

Temporarily, it uses a zk circuit which reveals to the Verifier the ranges being committed to. In the future, we will modify the circuit to hide the ranges.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "crates/components/hmac-sha256-circuits",
     "crates/components/key-exchange",
     "crates/components/poseidon-circomlib",
+    "crates/components/poseidon-bn256-pad14",
     "crates/components/stream-cipher",
     "crates/components/universal-hash",
     "crates/core",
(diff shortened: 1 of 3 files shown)


themighty1:
Re the above commit: use `poseidon-gadget`s permute; change to lbs0;

The above commit makes 2 sensitive changes to the circuit:
- uses a permutation chip from PSE's `poseidon-gadget` to achieve circomlib-compatible hashing
- handles plaintext bits in LSB0 order (before it was in MSB0) and counts the limbs from low to high. This is in line with how other projects deal with bits and limbs.
- introduces a zero advice cell to constrain values to be zero
- hashes the salt with rate-3 Poseidon, since rate-2 Poseidon of circomlib requires 57 partial rounds but `poseidon-gadget` supports only even amount of partial rounds.

PR #671, opened 2024-12-03
feat: poseidon-bn256-pad14

themighty1:
This PR adds a functionality to compute a Poseidon hash compatible with circomlib with custom input padding.

This is the ZK-friendly hash which will be used by the `tlsn-core` crate.

### DIFF
--- a/crates/components/poseidon-bn256-pad14/src/lib.rs
+++ b/crates/components/poseidon-bn256-pad14/src/lib.rs
@@ -1,0 +1,52 @@
+//! Poseidon hash over the bn256 curve with the input padding length of 14 field elements.
+
+use poseidon_circomlib::{hash as hash_inner, F};
+
+/// Maximum allowed bytelength of plaintext.
+const MAX_PLAINTEXT: usize = 434;
+/// How many bytes to pack into a single field element.
+const BYTES_PER_FIELD: usize = 31;
+/// The length to pad the plaintext field element count to.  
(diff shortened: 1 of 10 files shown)


crates/components/poseidon-bn256-pad14/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // Zero-pad if needed.
+    plaintext.extend(vec![F::zero(); PAD_LENGTH - plaintext.len()]);
+
+    plaintext.push(bytes_to_f(salt));
+
+    hash_inner(&plaintext)
+}
+
+/// Converts a little-endian byte representation of a scalar into a `F`.
```
yuroitaki, line 47: plaintext and salt passed into this crate are expected to be in little-endian format?

yuroitaki (commented):
nice! i think you have to update the base branch here as its not the latest yet?

crates/components/poseidon-bn256-pad14/src/lib.rs
yuroitaki: and hash() is returning the result in big endian? (due to `reverse()`)

themighty1: Good catch! 
I didn't think I needed to write a test for something as simple as what this crate does. It would have caught this error. Now I will add a test.

yuroitaki (approved):
LGTM!

PR #677, opened 2024-12-05
ci: reusable workflow fix & directory update

maceip:


maceip:
switched this PR to draft:

instead of triggering the SGX deployment automatically, will use the "click-to-deploy-via-actions-page" as seen in =>  #553

heeckhau:
Part of #686

Issue #541, opened 2024-07-24
question: async_tungstenite


0xJepsen:
https://github.com/tlsnotary/tlsn/blob/3554db83e17b2e5fc98293b397a2907b7f023496/notary/server/src/service/axum_websocket.rs#L1-L5

I am curious if would be better to use this fork of [`ws_stream_tungstenite`](https://github.com/pluto/ws_stream_tungstenite/commit/23a190d0edb3ca31fbd178e897faac67289d1503) and use `tokio_tungstenite` all the way up. The reason i am curious in this is that it seams to be the case to me that `tokio_tungstenite` has matured and become more of the standard and now has the support for the async properties that `async_tungstenite` initially added. These now seems to make a fully stale toolchain which `ws_stream_tungstenite` is a part of.

PR #681, opened 2025-01-02
chore: update `axum` to v0.8

heeckhau:


crates/notary/server/Cargo.toml
```diff
@@ -22,8 +22,8 @@ tlsn-verifier = { workspace = true }
 async-trait = { workspace = true }
```
yuroitaki, line 22: can remove this since it's no longer used

yuroitaki (changes requested):
LGTM! just one comment

crates/notary/server/src/service/axum_websocket.rs
```diff
@@ -584,17 +617,133 @@ impl Sink<Message> for WebSocket {
     }
 }
 
+/// UTF-8 wrapper for [Bytes].
+///
+/// An [Utf8Bytes] is always guaranteed to contain valid UTF-8.
```
yuroitaki, line 622:
```suggestion
/// An [Utf8Bytes] is always guaranteed to contain valid UTF-8.
/// The following NOTARY_MODIFICATION(s) are required because `async_tungstenite` (v0.28.2) is using an older version of `tungstenite` than `tokio_tungstenite` (v0.26.1). This older version of `tungstenite` (v0.26.0) doesn't have `Utf8Bytes`.
```

crates/notary/server/src/service/axum_websocket.rs
```diff
@@ -623,52 +772,72 @@ pub struct CloseFrame<'t> {
...
+    /// After receiving a close frame,
+    /// axum will automatically respond with a close frame if necessary
+    /// (you do not have to deal with this yourself).
+    /// Since no further messages will be received,
+    /// you may either do nothing
+    /// or explicitly drop the connection.
+    Close(Option<CloseFrame>),
 }
 
 impl Message {
```
yuroitaki, line 817:
```suggestion
/// The following NOTARY_MODIFICATION(s) are required because `async_tungstenite` (v0.28.2) is using an older version of `tungstenite` than `tokio_tungstenite` (v0.26.1).
impl Message {
```

yuroitaki (changes requested):
Good job in updating axum_websocket.rs! Though soon we definitely need to look into https://github.com/tlsnotary/tlsn/issues/541 to see if we can reduce the maintenance scope of this forked script

PR #683, opened 2025-01-15
Cipher crate

th4s:
This PR introduces a `cipher` crate using the new mpz alpha.1 VM. It supersedes `block-cipher` and `stream cipher` crates, which have been removed. It also removes `aead` crate as different encryption/decryption modes have been moved to tls-mpc crate (not part of this PR).


th4s:
:+1:

crates/components/cipher/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        if self.len() != len {
+            return Err(CipherError::new(format!(
+                "message has wrong length, got {}, but expected {}",
+                len,
+                self.len()
+            )));
+        }
+
+        let message_len = len as u32;
+        let block_count = (message_len / 16) + (message_len % 16 != 0) as u32;
```
themighty1, line 310: Did you mean to use
`let block_size = <<C as CipherCircuit>::Block as StaticSize<Binary>>::SIZE / 8;`
instead of hard-coding 16 ?

themighty1 (commented):
good work! 
just one q @th4s

crates/components/cipher/src/lib.rs
th4s: good catch :+1:  yeah that would have been the cleaner solution. But I think that code has been superseded by now.

PR #692, opened 2025-02-06
Feat/new encoder

th4s:


crates/core/src/transcript/encoding/encoder.rs
```diff
@@ -1,11 +1,72 @@
...
+impl ChaChaEncoder {
+    pub(crate) fn new(secret: &EncoderSecret) -> Self {
+        let seed = secret.to_seed();
+        let delta = u128::from_le_bytes(secret.to_delta());
+
+        Self { seed, delta }
+    }
+
+    pub(crate) fn new_prg(&self, stream_id: u64) -> ChaCha12Rng {
+        let mut prg = ChaCha12Rng::from_seed(self.seed);
```
themighty1, line 65: I don't recall us having a consensus to use chacha12 in prod. I thought we only used chacha**20** in prod. 
Besides, there was recent attacks on 8-round Salsa, https://eprint.iacr.org/2025/289

themighty1 (commented):
ack, but need to agree on using chacha20 @th4s  @sinui0

crates/core/src/transcript/encoding/encoder.rs
th4s: I wasn't aware of our consensus to only use `ChaCha20` in prod. My understanding was that `ChaCha12` is still considered secure (https://github.com/rust-random/rand/issues/932).

themighty1: nvm, we already use blake3 elsewhere in the code which is based on chacha but has only 7 rounds.
So, using chacha12 is justifiable.

PR #693, opened 2025-02-17
refactor: MPC-TLS

sinui0:
This PR effectively re-writes the MPC-TLS protocol to support changes to the `mpz` interface and includes additional optimizations around preprocessing and batching of TLS records.

An important change to be aware of is that the MPC-TLS implementation no longer handles committing to the plaintext of received data in the ZK VM. Only the tags are verified, and the protocol outputs the key references and the ciphertexts as the commitments. Committing and proving of the plaintext will be moved into the outer protocol.

Another change/simplification is the handling of different record types. The follower does not enforce any specific behavior in the record layer until _after the records are committed_. This helps simplify control flow everywhere and makes it easier to reason about what checks are made and when. This logic is consolidated into a single function `follower::validate_transcript`, and right now is quite strict.


themighty1:
ack

PR #696, opened 2025-02-25
refactor: prover + verifier

sinui0:
This PR refactors the Prover and Verifier to support newest changes to `mpz` and `mpc-tls`.


th4s (commented):
Looking good :+1:

PR #703, opened 2025-03-03
feat: harness

sinui0:
This PR consolidates our integration tests and benchmarks into a single harness.

# Highlights

- Shared code for tests and benches.
- Supports native and browser. This means that tests only have to be written once and they will be run in both environments.
- Simplify writing and maintaining tests using providers for dependency injection. We no longer have to duplicate setup and tear down of IO everywhere.
- Should reduce build/CI times.
- Clap CLI for configuration

### DIFF
--- a/crates/harness/runner/src/network.rs
+++ b/crates/harness/runner/src/network.rs
@@ -141,16 +141,8 @@ impl Network {
         ip_route(&self.ns_0, "default", &self.veth_rpc_0.0.name)?;
         ip_route(&self.ns_1, "default", &self.veth_rpc_1.0.name)?;
         ip_route(&self.ns_app, "default", &self.veth_app.0.name)?;
-        ip_route(
-            &self.ns_0,
-            self.config.proto_1.0,
-            &self.veth_proto_0.0.name,
-        )?;
-        ip_route(
-            &self.ns_1,
-            self.config.proto_0.0,
-            &self.veth_proto_1.0.name,
-        )?;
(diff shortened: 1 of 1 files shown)


crates/harness/executor/src/provider.rs
```diff
@@ -0,0 +1,115 @@
...
+                .map_err(anyhow::Error::from)
+        }
+
+        /// Provides a connection to the peer.
+        pub async fn provide_proto_io(&self) -> Result<impl Io> {
+            match self.mode {
+                IoMode::Client => {
+                    // It might take a bit for the peer to start up, so we retry a few times.
+                    let mut retries = 0;
+                    loop {
```
themighty1, line 50: Can we make sure that `TCP_NODELAY` is enabled for the TCP sockets.

crates/harness/runner/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+            benches.extend(items.to_benches(samples, samples_override));
+
+            runner.start_services().await?;
+            runner.exec_p.start().await?;
+            runner.exec_v.start().await?;
+
+            for config in benches {
+                runner
+                    .network
+                    .set_proto_config(config.bandwidth, config.protocol_latency.div_ceil(2))?;
```
themighty1, line 222: Isn't `protocol_latency`  == round-trip latency? why do we divide it by 2 here?

themighty1 (approved):
Nice work, ack

PR #706, opened 2025-03-03
chore: removed unused deps

sinui0:
This PR removes `mpz` dependencies from `tlsn-core` (yay), and one unused one from `tlsn-common`


th4s (commented):
:+1:

PR #731, opened 2025-03-12
Ethereum compatible signer

### LINKED ISSUE #694: secp256k1 signer is not Ethereum compatible
Our `secp256k1` signer we provide in `tlsn-core` is not compatible with Ethereum as the `k256` crate uses `sha256` to hash the message. We should update [our signer implementation](https://github.com/tlsnotary/tlsn/blob/4d5102b6e141ecb84b8a835604be1d285ae6eaa5/crates/core/src/signing.rs#L252) to hash the message using `keccak256`. See example [here](https://docs.rs/k256/latest/k256/ecdsa/index.html#computing-a-signature-with-a-recoveryid).

themighty1:
This PR adds an Ethereum compatibe signer, resolves #694

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
           components: clippy
 
       - name: Use caching
-        uses: Swatinem/rust-cache@v2.7.3
+        uses: Swatinem/rust-cache@v2.7.7
 
       - name: Clippy
         run: cargo clippy --keep-going --all-features --all-targets -- -D warnings
@@ -58,7 +58,7 @@ jobs:
           components: rustfmt
 
       - name: Use caching
-        uses: Swatinem/rust-cache@v2.7.3
+        uses: Swatinem/rust-cache@v2.7.7
 
       - name: Check formatting
         run: cargo +nightly fmt --check --all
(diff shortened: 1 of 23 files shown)


themighty1:
@sinui0 , I pushed a commit which fixes an incorrect recovery id. But even with the fix, the test is not passing against the data you provided, the sigs are completely different.

```rust
#[test]
    fn test_secp256k1eth() {
        // Test vector obtained by hacking ecrecover.
        let priv_key =
            decode("f219198685f5ab3a4b7a675cf18f7048408e8348a77ec2a1a02c75b9d965fbea").unwrap();
        let expected_sig =  decode("081c81ec33f1cc351a0415e4f457f4020862fa8dca8f4435e8a8e5e92a16a6e926f42b5eefc2e828e8926b59cb24b50661a632e0079ab1a69277e8142673b7ee1c").unwrap();
        let msg = "test message";

        let signer: Secp256k1EthSigner = Secp256k1EthSigner::new(&priv_key).unwrap();
        let sig = signer.sign(msg.as_bytes()).unwrap();

        assert_eq!(sig.data, expected_sig);
    }
    ```

themighty1:
@sinui0, testing against alloy-rs now. the test passes ---^

PR #733, opened 2025-03-14
fix: record layer handshake control flow

### LINKED ISSUE #729: bug: record layer flushing in wrong state
A user has reported this error in the alpha.8 release:

```sh
Starting an MPC TLS connection with the server
2025-03-11T21:55:49.163183Z ERROR mpc_tls::leader: error=internal error: "state error: must be in active or closed state to flush record layer"
2025-03-11T21:55:49.217332Z ERROR try_tlsn_alpha_8: connection closed before message completed
```

I suspect this could be caused by the client trying to flush the record layer before the handshake has been completed. Simplest fix is to emit a warning log instead of returning an error if this occurs. Better fix is to ensure the client does not try to flush the record layer at the wrong time.

sinui0:
Closes #729 

This PR adjusts MPC-TLS to allow record buffering prior to handshake completion. We no longer return an error if flush is called prior to handshake completion. The flush implementation was also changed so that application data gets processed immediately but will not be available until the client has signaled to start traffic.

### DIFF
--- a/crates/common/src/transcript.rs
+++ b/crates/common/src/transcript.rs
@@ -167,3 +167,71 @@ impl TranscriptRefs {
 #[derive(Debug, thiserror::Error)]
 #[error("not all application plaintext was committed to in the TLS transcript")]
 pub struct IncompleteTranscript {}
+
+#[cfg(test)]
+mod tests {
+    use super::TranscriptRefs;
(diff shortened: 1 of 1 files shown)


th4s (approved):

themighty1:
ack

PR #744, opened 2025-03-25
doc: add comments in example for commit and reveal

yuroitaki:


crates/examples/attestation/present.rs
```diff
@@ -64,10 +66,12 @@ async fn create_presentation(example_type: &ExampleType) -> Result<(), Box<dyn s
...
     let response = &transcript.responses[0];
+    // Reveal the structure of the response without the headers or body.
     builder.reveal_recv(&response.without_data())?;
-    for header in &response.headers {
-        builder.reveal_recv(header)?;
-    }
+    // Reveal all response headers.
+    let start = response.headers.first().unwrap().to_range_set().min().unwrap();
+    let end = response.headers.last().unwrap().to_range_set().end().unwrap();
+    builder.reveal_recv(&(start..end))?;
```
heeckhau, line 74: The for loop was simpler. Why this change?

yuroitaki: i wanted to showcase the new feature where one can reveal once for multiple committed ranges (https://github.com/tlsnotary/tlsn/issues/655) — imo wouldn't obvious if we don't have it in the example.

alternatively we can document it in our doc page and link it here if we need the simplicity in example?

yuroitaki: > I don't think we need to showcase this new behavior, it is the default behavior that users will be expecting. This change adds noise to the example

is that cause the `reveal` APIs take rangeset as argument, hence it's fair to assume users will naturally expect that they can union multiple ranges and reveal once?

yuroitaki:
> Closing, see comments

Gotcha, though there is a minor fix (removing commented line) and some comments I want to include, so opened https://github.com/tlsnotary/tlsn/pull/761 for them

PR #746, opened 2025-03-25
Log notarization elapsed time

LogvinovLeon:
We would like to track - how much time did notarization take. Now - we need to find the start log, find the end log and subtract times. After this PR - it becomes straight-forward

### DIFF
--- a/crates/notary/server/src/service/tcp.rs
+++ b/crates/notary/server/src/service/tcp.rs
@@ -6,6 +6,7 @@ use axum::{
 use axum_core::body::Body;
 use hyper::upgrade::{OnUpgrade, Upgraded};
 use hyper_util::rt::TokioIo;
+use tokio::time::Instant;
 use std::future::Future;
 use tracing::{debug, error, info};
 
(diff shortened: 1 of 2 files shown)


crates/notary/server/src/service/tcp.rs
```diff
@@ -84,13 +85,22 @@ pub async fn tcp_notarize(
...
     session_id: String,
 ) {
+    let start = Instant::now();
     debug!(?session_id, "Upgraded to tcp connection");
     match notary_service(stream, notary_globals, &session_id).await {
         Ok(_) => {
-            info!(?session_id, "Successful notarization using tcp!");
+            info!(
+                ?session_id,
+                elapsed_time = start.elapsed().as_millis(),
```
yuroitaki, line 94: can we have the unit (milliseconds) printed too?

yuroitaki (commented):
Nice work! Got a comment

crates/notary/server/src/service/tcp.rs
LogvinovLeon: I didn't want to include time in message so that I can later filter my message.
Therefore - changed the field name to include the units:
```
Successful notarization using tcp! session_id="..." elapsed_time_millis=0
```

PR #747, opened 2025-03-25
fix comments

themighty1:
This PR improves comment grammar and clarity.


crates/mpc-tls/src/record_layer/aead/aes_gcm.rs
```diff
@@ -375,7 +377,6 @@ impl MpcAesGcm {
     /// # Arguments
     ///
     /// * `vm` - Virtual machine.
-    /// * `inputs` - Data to verify the tags for.
```
th4s, line 378: The argument description does not match the function arguments.

crates/mpc-tls/src/record_layer/decrypt.rs
```diff
@@ -197,7 +197,9 @@ impl DecryptOp {
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub(crate) enum DecryptMode {
+    /// The decrypted plaintext is private.
     Private,
+    /// The decrypted plaintext is public.
```
th4s, line 202: shouldn't this be `the decrypted ciphertext`?

th4s (approved):

crates/mpc-tls/src/record_layer/decrypt.rs
themighty1: will just use `plaintext` here

PR #748, opened 2025-03-25
fix: avoid mutating self in TagShare::add

themighty1:
This PR removes `mut self` in TagShare::add since the mutability is not required.

I checked all references to `TagShare` in the code and couldnt find where it needs to be mutable.


th4s (approved):

themighty1:
> Please base on `alpha.3` so we avoid conflicts later

This is a PR in `tlsn`, I don't see an `alpha.10` branch, so we merge into `dev`, right? @th4s , could you confirm?

th4s:
yes, you can merge onto dev. I will rebase my branches.

PR #751, opened 2025-03-26
ci: SGX build: drop TEE GH environment, use regular secret

heeckhau:


.github/scripts/gramine.sh
```diff
@@ -1,43 +0,0 @@
-#/bin/sh
```
yuroitaki, line 1: what about the `cd-scripts` folder, do we still need those?

yuroitaki (approved):
just one comment

PR #754, opened 2025-03-26
feat(mpc-tls): improve error message for incorrect transcript config

sinui0:
This PR improves the error messaging for when the transcript exceeds what was configured.


th4s (approved):
looking good, but needs to be rebased.

PR #770, opened 2025-04-09
feat(notary): add concurrency limit

themighty1:
This PR adds adds a concurrency limit feature to the notary server allowing it to limit the number of concurrent notarizations.

This PR does not implement any queueing mechanism. Instead the client retries the request until it is accepted.
There are many ways to implement queueing with different trade-offs and I left it out of scope for this PR.


themighty1:
@sinui0 , comments addressed but the exact strategy for retries e.g. exponential backoff I decided to leave out of scope for this PR.

themighty1:
@sinui0, all comments addressed

Issue #739, opened 2025-03-20
Expose record count configuration in tlsn_common::config::ProtocolConfig


sinui0:
We need to give the user the options to override the estimate for the TLS record count in the MPC-TLS preprocessing.

PR #772, opened 2025-04-11
feat(mpc-tls): add config to recv record cushion

themighty1:
This PR adds a config option for a cushion to handle cases when the number of incoming TLS record is underestimated by the internal algorithm.

Closes https://github.com/tlsnotary/tlsn/issues/739

### DIFF
--- a/crates/common/src/config.rs
+++ b/crates/common/src/config.rs
@@ -27,6 +27,15 @@ pub struct ProtocolConfig {
     max_recv_data_online: usize,
     /// Maximum number of bytes that can be received.
     max_recv_data: usize,
+    /// A cushion for handling more incoming TLS records than estimated by the
+    /// internal algorithm.
+    ///
+    /// We internally estimate based on an optimistic assumption that the
+    /// incoming TLS record's size will be 16KB, but in practice it is often
+    /// not the case.
+    /// This value will be added to the internal estimate.
+    #[builder(default = "0")]
+    extra_recv_records: usize,
     /// Version that is being run by prover/verifier.
     #[builder(setter(skip), default = "VERSION.clone()")]
     version: Version,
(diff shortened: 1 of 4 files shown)


th4s (commented):
Hmm I think I would have chosen a different approach. 

- Add fields `max_sent_records` and `max_recv_records` to `ProtocolConfig` that are optional
- In `ProtocolConfigValidator` add a reasonable check to `validate` for these fields
- Then in `build_mpc_tls_config` just forward these values if they are set.

I think this way it is cleaner and we would also take `max_sent_records` into account.

themighty1:
> I think this way it is cleaner and we would also take `max_sent_records` into account.
I was operating under the assumption that the user does not need control over the **sent** records since we will never run into issues with underestimating the **sent** records count. Or will we run into an issue? I can't think of when.

> Add fields `max_sent_records` and `max_recv_records` to `ProtocolConfig` that are optional

Allowing the user to compute `max_recv_records` on their own means they have to be additionally exposed to `MIN_RECV_RECORDS` from crates/mpc-tls/src/config.rs. Essentially they have to understand how many TLS records to allocate for the handshake.

I chose the path which leaks as few details as possible to the user.

But, frankly, I feel that this feature is a temporary crutch and it will go away in favor of computing `j0`s dynamically during the session.

th4s:
I see, what is the reason we never run into issues with underestimating the sent record count? I thought this is similar to the received record count?

> Allowing the user to compute max_recv_records on their own means they have to be additionally exposed...

I see, this makes sense. So instead of specifying the exact record count the user can specify the extra amount.

themighty1:
> I see, what is the reason we never run into issues with underestimating the sent record count?

I can't show the exact place in the code but Im sure that rustls always chunks the outbound data into the largest possible 16KB TLS records, hence we don' run into estimation issues.

@th4s , could you ACK, so I can merge?

PR #776, opened 2025-04-14
misc(notary): improve error msg when tls is expected

themighty1:
This PR improves the error message in cases when the client or notary unexpectedly get a connection without TLS enabled.

### DIFF
--- a/crates/notary/client/src/client.rs
+++ b/crates/notary/client/src/client.rs
@@ -180,23 +180,9 @@ impl NotaryClient {
                 )
                 .await
                 .map_err(|err| {
-                    if let Some(rustls::Error::InvalidMessage(
-                        rustls::InvalidMessage::InvalidContentType,
-                    )) = err
-                        .get_ref()
-                        .and_then(|inner| inner.downcast_ref::<rustls::Error>())
-                    {
+                    if is_tls_mismatch_error(&err) {
                         error!("Perhaps the notary server is not accepting our TLS connection");
-                    } else if let Some(err) = err
-                        .get_ref()
-                        .and_then(|inner| inner.downcast_ref::<hyper::Error>())
-                    {
-                        // Check if the hyper error is `Kind::Parse(Parse::Version)`.
-                        if err.to_string() == "invalid HTTP version parsed" {
-                            error!("Perhaps the notary server is not accepting our TLS connection");
-                        }
                     }
-
                     ClientError::new(ErrorKind::TlsSetup, Some(Box::new(err)))
                 })?;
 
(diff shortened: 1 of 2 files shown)


crates/notary/client/src/client.rs
```diff
@@ -179,7 +179,23 @@ impl NotaryClient {
                     notary_socket,
                 )
                 .await
-                .map_err(|err| ClientError::new(ErrorKind::TlsSetup, Some(Box::new(err))))?;
+                .map_err(|err| {
+                    if let Some(inner) = err.get_ref() {
```
yuroitaki, line 183: ` { kind: Http, source: Some(hyper::Error(Parse(Version))) }` could also indicate tls connection mismatch issue — where client tries to send non-TLS connection, and server is expecting it — i think we should also add that

yuroitaki: though both `Parse(Version)` and `InvalidContentType` doesn't only mean TLS connection mismatch, but could arise from other issues based on their type definitions; 

hence i think our error message shouldnt be so affirmative, but a more conservative suggestion might be better; e.g. `it's possible that notary server is not accepting our TLS connection`

crates/notary/server/src/server.rs
```diff
@@ -210,6 +210,15 @@ pub async fn run_server(config: &NotaryServerProperties) -> Result<(), NotarySer
                     }
                     Err(err) => {
                         error!("{}", NotaryServerError::Connection(err.to_string()));
+                        if let Ok(error) = err.downcast::<rustls::Error>() {
+                            if error
+                                == rustls::Error::InvalidMessage(
+                                    rustls::InvalidMessage::InvalidContentType,
+                                )
```
yuroitaki, line 217:
```
  if let Some(rustls::Error::InvalidMessage(rustls::InvalidMessage::InvalidContentType)) =
       err.get_ref().and_then(|inner| inner.downcast_ref::<rustls::Error>())
```

this could be neater

crates/notary/server/src/server.rs
```diff
@@ -210,6 +210,15 @@ pub async fn run_server(config: &NotaryServerProperties) -> Result<(), NotarySer
...
+                            {
+                                error!("You should turn on `TLSProperties::enabled` to accept client's TLS connections");
```
yuroitaki, line 219: same comment as above on the wording

yuroitaki (changes requested):
gw! some comments

themighty1:
all comments addressed @yuroitaki

crates/notary/client/src/client.rs
```diff
@@ -485,6 +474,26 @@ fn default_root_store() -> RootCertStore {
...
+        .get_ref()
+        .and_then(|inner| inner.downcast_ref::<rustls::Error>())
+    {
+        return true;
+    } else if let Some(err) = err
+        .get_ref()
+        .and_then(|inner| inner.downcast_ref::<hyper::Error>())
+    {
+        // Check if the hyper error is `Kind::Parse(Parse::Version)`.
+        if err.to_string() == "invalid HTTP version parsed" {
```
yuroitaki, line 490: Sorry I just realised that `Kind::Parse(Parse::Version)` is not triggered at [`request_notarization()`](https://github.com/tlsnotary/tlsn/blob/de7a47de5b4f6e4de0ec2ff3d1c42a0d2bf7a6a6/crates/notary/client/src/client.rs#L187); rather at [`send_request()`](https://github.com/tlsnotary/tlsn/blob/de7a47de5b4f6e4de0ec2ff3d1c42a0d2bf7a6a6/crates/notary/client/src/client.rs#L298)

yuroitaki: Anyway that doesn't matter, since these error messages shouldn't be relied upon according to [hyper's doc](https://github.com/hyperium/hyper/blob/0369f5e86406abbd94091ee5120181632c5d98d2/src/error.rs#L20-L22). The only other way I know is
```rust
if format!("{:?}", err).contains("Parse(Version)")
```
which also is kinda brittle.

Given that, I am now leaning towards not catching this error in the client side anymore, since it will be caught in notary server. Let me know what you think.

yuroitaki (changes requested):

themighty1:
good catch on the wrong error msg, all feedback addressed, thanks!

yuroitaki (commented):
ack

PR #780, opened 2025-04-15
chore: improve naming and comments

themighty1:
This PR makes no functional changes, just improves naming and comments and addresses @th4s 's comments raised in https://github.com/tlsnotary/tlsn/pull/696

### DIFF
--- a/crates/common/src/encoding.rs
+++ b/crates/common/src/encoding.rs
@@ -19,6 +19,8 @@ struct Encodings {
 }
 
 /// Transfers the encodings using the provided seed and keys.
+///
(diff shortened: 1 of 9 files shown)


crates/common/src/encoding.rs
```diff
@@ -132,7 +136,7 @@ impl EncodingProvider for Provider {
             Direction::Received => &self.recv,
         };
 
-        let mut encoding = Vec::with_capacity(idx.len());
+        let mut encoding = Vec::with_capacity(idx.len() * ENCODING_SIZE);
```
themighty1, line 139: I guess this one can be considered the only functional change in this PR.

PR #781, opened 2025-04-15
perf(core): simplify encoding logic

themighty1:
This PR simplifies the encoding logic.


themighty1:
Oops, let me make it constant time, it will still be cleaner than the original code.

themighty1:
@sinui0 , ready

th4s (approved):

PR #783, opened 2025-04-17
fix: select non-app recs correctly

themighty1:
This PR fixes a bug where non-application data records were incorrectly selected.
Also makes sure that the received records are not sorted by the order of decryption (which happens when deferred decryption is in use) but instead by their sequence number.


themighty1:
@sinui0  Before my change if `is_deferred == true` and `decrypt_buffer` contains `(ApplicationData, Alert(Close_Notify))`,  the code would **not** decrypt the Alert in MPC.

We want the Alert to be decrypted in MPC, right?

themighty1:
nvm, I guess I was confused by the nomenclature `decrypt::decrypt_local` . Upon a careful look I see that there is some MPC involved in that fn for public msgs. That's where the Alert will be decrypted.

PR #786, opened 2025-04-18
feat: expose record count config

### LINKED ISSUE #739: Expose record count configuration in tlsn_common::config::ProtocolConfig

sinui0:
Closes #739 

This exposes the record count config to the user so they can override it if necessary. It also adjusts the default count estimation to be more liberal in its estimation.

### DIFF
--- a/.github/workflows/releng.yml
+++ b/.github/workflows/releng.yml
@@ -6,7 +6,7 @@ on:
       tag:
         description: 'Tag to publish to NPM'
         required: true
-        default: 'v0.1.0-alpha.9'
+        default: 'v0.1.0-alpha.10'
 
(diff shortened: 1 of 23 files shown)


themighty1 (approved):
aqk

th4s (commented):
looking good, one thing:

What about the wasm config: https://github.com/tlsnotary/tlsn/blob/main/crates/wasm/src/prover/config.rs#L7

themighty1:
Good catch @th4s , I'll work on https://github.com/tlsnotary/tlsn/issues/799

PR #795, opened 2025-04-22
chore: add rustls licence and attribution

### LINKED ISSUE #299: Add Rustls Licenses to `tls-client` and `tls-core`
Give proper attribution to `rustls` in both crates.

themighty1:
This PR adds the missing rustls licence and attribution.

Closes #299


heeckhau:
@themighty1 are these the only crates that have rustls code?

themighty1:
@heeckhau , yes, only those two

PR #805, opened 2025-04-23
Build gramine-sgx for dev and tagged builds only

heeckhau:


yuroitaki (commented):
do we still need `build-sgx` to run for every PR then? or also only for dev and tag?

PR #818, opened 2025-04-29
refactor(notary): remove config file and fixtures

yuroitaki:
https://github.com/tlsnotary/tlsn/issues/773, https://github.com/tlsnotary/tlsn/issues/766

This PR includes:
- removing compulsory config.yaml and having defaults in code
- supporting overrides via env var or optional config file
- prepending any relative paths in optional config file
- restructuring different internal modules for better scope division/clarity

This PR excludes the following to avoid becoming too huge for reviews:
- default to ephemeral notary key gen
- fixtures removal
- any documentation/CI/docker updates

Instead, they will be in separate upcoming PRs against this (DO NOT MERGE this before they are merged in first).

PR #819, opened 2025-04-30
Fixture removal

yuroitaki:
- generate signing key if not provided
- migrate fixture to test crate
- removes redundant code from tee module
- derive pub key from provided signing key


crates/notary/server/src/signing.rs
```diff
@@ -72,3 +102,73 @@ enum SigningKey {
...
+        let public_key_bytes = self.to_compressed_bytes();
+        let subject_public_key = der::asn1::BitStringRef::new(0, &public_key_bytes)?;
+
+        pkcs8::SubjectPublicKeyInfo {
+            algorithm,
+            subject_public_key,
+        }
+        .try_into()
+    }
+}
```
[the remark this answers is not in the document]
yuroitaki: Agreed, didn't realise technically all of `PublicKey` (and `VerifyingKey`) functionalities are already present in `Signer` except for `to_pem`; so if that's added, then we don't need `Publickey` anymore since in notary server crate I can just do
- `signer.verifying_key()` 
- `signer.to_pem()`

right?

crates/notary/server/src/server.rs
```diff
@@ -246,26 +249,31 @@ fn build_crypto_provider(attestation_key: AttestationKey) -> CryptoProvider {
...
+            .map_err(|err| eyre!("Failed to load notary signing key for notarization: {err}"))?;
 
-    pem.zeroize();
+        pem.zeroize();
 
-    debug!("Successfully loaded notary server's signing key!");
+        key
+    } else {
+        warn!(
+            "WARNING! A *random* (ephemeral) signing key will be generated, since none is provided"
```
heeckhau, line 272: `warn!` already makes this a warning. I also suggest this to make it more in the style of the other messges.

```suggestion
            "⚠️ Using a random, ephemeral signing key because `private_key_path` is not set."
```

heeckhau:
@yuroitaki The TEE config needs an update too. Or should we drop it?

`crates/notary/server/tee/config/config.yaml`

```
  TOML parse error at line 1, column 7
    |
  1 | server:
    |       ^
  expected `.`, `=`
  ```

File: crates/notary/server/src/signing.rs [hunk 3]
```diff
@@ -47,6 +50,25 @@ impl TryFrom<PrivateKeyInfo<'_>> for AttestationKey {
...
+                    &mut rand::rng().compat(),
+                )),
+            }),
+            "SECP256R1" => Ok(Self {
+                alg_id: SignatureAlgId::SECP256R1,
+                key: SigningKey::Secp256r1(p256::ecdsa::SigningKey::random(
+                    &mut rand::rng().compat(),
+                )),
+            }),
+            alg_id => Err(eyre!("unsupported signature algorithm: {alg_id} — only secp256k1 and secp256r1 are supported now")),
```
heeckhau, line 68:
```suggestion
            alg_id => Err(eyre!("unsupported signature algorithm: {alg_id} — only secp256k1 and secp256r1 are supported")),
```

heeckhau (commented):
Some remarks:
* We should not forget add a sample config file in the docs, so that users can easily modify it
* the server config that is printed at startup is not a valid config:
   * `Compact` -> `COMPACT`
   * `html_info:` should have a trailing `|` 
* How does one create a config that uses ephemeral keys?
* Is the TEE config still valid?

yuroitaki:
> Some remarks:
> 
> * We should not forget add a sample config file in the docs, so that users can easily modify it
> * the server config that is printed at startup is not a valid config:
>   
>   * `Compact` -> `COMPACT`
>   * `html_info:` should have a trailing `|`
> * How does one create a config that uses ephemeral keys?
> * Is the TEE config still valid?

will address these in https://github.com/tlsnotary/tlsn/pull/818

PR #875, opened 2025-05-20
refactor(core): decouple attestation from core api

sinui0:
This PR decouples the attestation code from the core API. In a following PR we can fully extract attestation types into a separate crate.

# Changes
- Transcript commitment is no longer coupled to the attestation request.
- Notarize methods are no longer a fork in the (prover/verifier) APIs, but instead use the core API directly. I've also marked them as deprecated to signal to users that they will be removed.
- Prover/Verifier is now more declarative using the `(Prove/Verify)Config` types.
- Encoding commitment protocol is further encapsulated in `tlsn-common`.


th4s (approved):
lgtm

PR #881, opened 2025-05-22
feat: SHA256 transcript commitments

### LINKED ISSUE #674: feat: Support SHA256 transcript commitments

sinui0:
Closes #674 

This PR adds support for hash commitments to transcript plaintext, and starts with supporting SHA256.

With this, it will now be possible for the Prover to prove blinded hashes of arbitrary subsets of the plaintext data to the Verifier. However, this PR also adds an error condition which prevents the Prover from requesting a hash commitment in an attestation which does not cover the entire transcript. This is to preserve the obliviousness of the reference notary implementation. This constraint will be relaxed in the future, or at least there will be alternative options to work around it. **This constraint only applies to the attestation flow involving a Notary. No such constraint is imposed between a Prover and Verifier.**

### DIFF
--- a/crates/common/src/commit/hash.rs
+++ b/crates/common/src/commit/hash.rs
@@ -22,6 +22,7 @@ use crate::{transcript::TranscriptRefs, Role};
 /// Future which will resolve to the committed hash values.
 #[derive(Debug)]
 pub struct HashCommitFuture {
+    #[allow(clippy::type_complexity)]
     futs: Vec<(
         Direction,
         Idx,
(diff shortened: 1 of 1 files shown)


crates/verifier/src/lib.rs
th4s, line 1: What about also restricting the verifier when doing a notarization to only allow a hash commitment over the full transcript for now?

th4s (approved):
looking good :rocket:, some non-blocking remarks

PR #888, opened 2025-05-23
Build wasm with --release mode

heeckhau:
I noticed these warnings on CI:
```
[WARN]: ⚠️   "package.metadata.wasm-pack.profile.debug" is an unknown key and will be ignored. Please check your Cargo.toml.
[WARN]: ⚠️   "package.metadata.wasm-pack.profile.release.?.opt-level" is an unknown key and will be ignored. Please check your Cargo.toml.
```
Turns out these settings no longer work, and we should build in release mode instead


heeckhau:
Dropped the profiles because of:
```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/runner/work/tlsn/tlsn/crates/benches/browser/wasm/Cargo.toml
workspace: /home/runner/work/tlsn/tlsn/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/runner/work/tlsn/tlsn/crates/wasm/Cargo.toml
workspace: /home/runner/work/tlsn/tlsn/Cargo.toml
```

heeckhau:
@sinui0 @themighty1 

FYI:
* `--release` is added by default to `wasm-pack build` but I have added it explicitly to make this clear
* `--wasm-opt` is also default (you need to add `no-opt` explicitly to skip it)
* debug was deprecated and was renamed to dev

PR #903, opened 2025-06-03
feat: add logging to server fixture (#901)

### LINKED ISSUE #901: App server fixture logging
Configure logging for the app server fixture so the user can see what's happening while running the examples.

heeckhau:
fixes #901

### DIFF
--- a/crates/examples/interactive/README.md
+++ b/crates/examples/interactive/README.md
@@ -4,7 +4,7 @@ This example demonstrates how to use TLSNotary in a simple interactive session b
 
 This example fetches data from a local test server. To start this server, run:
 ```shell
-PORT=4000 cargo run --bin tlsn-server-fixture
+RUST_LOG=info PORT=4000 cargo run --bin tlsn-server-fixture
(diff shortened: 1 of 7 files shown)


crates/examples/interactive/README.md
```diff
@@ -4,7 +4,7 @@ This example demonstrates how to use TLSNotary in a simple interactive session b
 
 This example fetches data from a local test server. To start this server, run:
 ```shell
-PORT=4000 cargo run --bin tlsn-server-fixture
+RUST_LOG=info PORT=4000 cargo run --bin tlsn-server-fixture
```
yuroitaki, line 7: why do we need to run this at port 4000 again? i always just use the default port 3000

yuroitaki (approved):
LGTM

crates/examples/interactive/README.md
heeckhau: Port 3000 is also the default port for the React development server. So this often conflicts.

PR #904, opened 2025-06-03
docs: improve example readme

heeckhau:


yuroitaki (approved):
TIL we can also use `--package` to run crate A binary in crate B folder; e.g. `cd crates/examples/; cargo run --package tlsn-server-fixture --bin tlsn-server-fixture`

https://users.rust-lang.org/t/cargo-run-bin-on-subcrate/108288

Though I think asking user to run at root level is probably as easy too

PR #907, opened 2025-06-05
refactor(mpc-tls): remove commit-reveal from tag verification

### LINKED ISSUE #889: Simplify tag verification in MPC-TLS
Currently our 2PC-GHASH protocol is asymmetric. For computing a tag for a ciphertext that is being sent to the TLS server, the Verifier can simply send their tag share to the Prover because this is sufficient to ensure the Prover can not manipulate the ciphertext prior to sending it. In such a case, the TLS server would reject the ciphertext because the Prover can not authenticate an alternate.

For tag verification, the Prover must first commit to their tag share prior to learning the Verifier's share otherwise the Prover could trivially adjust their share to construct a valid tag for a manipulated ciphertext. However, this is only necessary if we want to authenticate the ciphertext for the Verifier. Given that we are moving to a more efficient approach (#678) for authenticating tags for the Verifier, we can relax the 2PC-GHASH protocol by removing the commit-reveal. This will reduce the round complexity of the online phase.

sinui0:
Closes #889 

This PR removes the commit-reveal process from tag verification, which as of #868 it is no longer needed to authenticate the ciphertext for the verifier during the online phase.

### DIFF
--- a/crates/mpc-tls/src/record_layer.rs
+++ b/crates/mpc-tls/src/record_layer.rs
@@ -450,16 +450,15 @@ impl RecordLayer {
         let verify_tags = decrypt::verify_tags(&mut (*vm), &mut decrypter, &decrypt_ops)?;
 
         // Run tag computation and VM in parallel.
-        let (mut tags, _, _) = ctx
-            .try_join3(
-                async move |ctx| {
-                    compute_tags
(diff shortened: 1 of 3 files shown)


themighty1 (approved):
ack

PR #908, opened 2025-06-05
feat: peer error

sinui0:
This PR adds a peer error type which can be used to communicate error conditons between parties when acceptable. To start, the Verifier will let the Prover know why it is rejecting a protocol configuration.

### DIFF
--- a/crates/verifier/src/lib.rs
+++ b/crates/verifier/src/lib.rs
@@ -118,6 +118,9 @@ impl Verifier<state::Initialized> {
                 )
                 .await?;
 
+            mux_ctrl.close();
+            mux_fut.await?;
+
             return Err(e.into());
         }
 


yuroitaki (commented):
Looks good! Though in my machine, the attestation prover doesn't seem to print the `PeerError`, but the `io` error

image

image

crates/verifier/src/lib.rs
```diff
@@ -103,16 +104,25 @@ impl Verifier<state::Initialized> {
...
+                    ctx.io_mut()
+                        .send::<Result<(), PeerError>>(Err(PeerError::from(&e))),
+                )
+                .await?;
+
+            mux_ctrl.close();
+            mux_fut.await?;
+
+            return Err(e.into());
+        }
```
yuroitaki, line 125: we also need to send `Ok(())` to prover if validation succeeds right? now when I run the example where validation passed, prover just hung there because it's waiting for the validation result

PR #911, opened 2025-06-06
fix: set TCP_NODELAY for prover and notary

themighty1:
This PR sets `TCP_NODELAY` on prover and notary TCP sockets in order to improve the protocol latency.


crates/notary/client/src/client.rs
```diff
@@ -189,6 +189,10 @@ impl NotaryClient {
             let notary_socket = tokio::net::TcpStream::connect((self.host.as_str(), self.port))
                 .await
                 .map_err(|err| ClientError::new(ErrorKind::Connection, Some(Box::new(err))))?;
+            // Setting TCP_NODELAY will improve prover latency.
+            notary_socket
```
yuroitaki, line 193: this should also be configured for when client has TLS disabled right?

Issue #775, opened 2025-04-14
Support notarization for sites requiring Client Certificates (mTLS)


mkalka:
**Description:**
TLSNotary (TLSN) currently supports notarizing TLS 1.2 sessions, but it can't handle sessions requiring mutual TLS authentication (mTLS). Mutual TLS involves both client and server presenting certificates to authenticate each other during the TLS handshake.

Specifically, the feature would need to support:

- Inclusion of the client certificate (X.509) during the TLS 1.2 handshake phase.

- Proper handling of optional TLS 1.2 handshake messages involving client certificate (`CertificateRequest`, `Certificate`, and `CertificateVerify`)

**Context & Use-case:**

Adding this capability would broaden TLSN's usability in sectors relying heavily on mutual authentication eg.

- Proving interactions with financial APIs requiring client certificates.

- Notarizing government portals and regulated services which enforce mTLS.

PR #916, opened 2025-06-10
feat(prover): client authentication

themighty1:
This PR enables the prover to use TLS client authentication.
Also updates the examples to show how to use it.

Closes https://github.com/tlsnotary/tlsn/issues/775


themighty1:
ready for review @sinui0

PR #925, opened 2025-06-16
fix(harness): iptable rule and bench config variable

sinui0:
This PR adds an `iptables` forwarding rule which fixes an issue when the system already has a forwarding drop rule in place (eg. docker is installed). I also fixed a naming issue in the bench config.


heeckhau (approved):
Lgtm

PR #933, opened 2025-07-01
build(wasm): add wasm profile and optimize for perf

sinui0:
This PR makes a couple changes to our build process for WASM.

1. Define a `wasm` build profile instead of using package specific config. The build profile is used by both the harness and wasm packages.
2. Configure `wasm-opt` to optimize for speed

### DIFF
--- a/crates/harness/build.wasm.sh
+++ b/crates/harness/build.wasm.sh
@@ -4,4 +4,8 @@
 cd "$(dirname "$0")"
 
 rustup run nightly \
-    wasm-pack build executor --target web --no-pack --out-dir=../static/generated -- -Zbuild-std=panic_abort,std
+    wasm-pack build executor \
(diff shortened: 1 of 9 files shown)


crates/harness/build.wasm.sh
```diff
@@ -4,4 +4,8 @@
 cd "$(dirname "$0")"
 
 rustup run nightly \
-    wasm-pack build executor --target web --no-pack --out-dir=../static/generated -- -Zbuild-std=panic_abort,std
\ No newline at end of file
+    wasm-pack build executor \
+        --profile wasm \
+        --target web \
+        --out-dir=../static/generated \
+        -- -Zbuild-std=panic_abort,std
```
th4s, line 11: The `--no-pack` argument is missing. Is this on purpose?

PR #952, opened 2025-07-30
Remove notarize methods for prover and verifier

th4s:
This PR removes the deprecated `notarize` and `notarize_with_provider` methods from the prover and verifier.


crates/tlsn/src/verifier.rs
```diff
@@ -152,59 +152,6 @@ impl Verifier<state::Initialized> {
         })
```
yuroitaki, line 152: what about the other `notarize_with_provider` and `notarize` methods in `Verifier<state::Committed>`?

PR #977, opened 2025-08-27
Reduce WASM size with wasm profile settings

heeckhau:
This PR reduces the generated Wasm size by ~9% (from 9,266,036 to 8,444,117 bytes) by adjusting the wasm profile settings.

Notes
* Benchmarks have not been run, but I expect no negative performance impact
* Setting `lto = "fat"` was tested: it increases build time but does not produce a meaningful size improvement compared to lto = true.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,8 @@ opt-level = 1
 [profile.wasm]
 inherits = "release"
 lto = true
+panic = "abort"
+codegen-units = 1
 
 [workspace.dependencies]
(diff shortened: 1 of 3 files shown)


themighty1:
I benched before and after this PR and found no difference:

[before.csv](https://github.com/user-attachments/files/22042182/before.csv)
[after.csv](https://github.com/user-attachments/files/22042185/after.csv)

themighty1 (approved):

themighty1:
not only it is faster in every instance, we also reduce the size, should be goof to merge.

Issue #802, opened 2025-04-22
Improve error from ZkAesCtr caused by insufficient transcript config


sinui0:
After MPC-TLS we use `ZkAesCtr` to commit to the response plaintext. If the transcript config was set too low it will return a cryptic error about insufficient preprocessing. Improve this error message for users.

PR #1003, opened 2025-09-30
chore: improve error msg

themighty1:
This PR improves an error message when proving received plaintext.

fixes: https://github.com/tlsnotary/tlsn/issues/802

### DIFF
--- a/crates/tlsn/src/commit.rs
+++ b/crates/tlsn/src/commit.rs
@@ -91,6 +91,13 @@ impl RecordProofError {
     {
         Self(ErrorRepr::Vm(err.into()))
     }
+
+    pub(crate) fn is_insufficient(&self) -> bool {
+        match &self.0 {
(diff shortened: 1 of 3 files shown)


themighty1:
> I feel like we can reach the same goal by just improving the error message here:
> 
> https://github.com/tlsnotary/tlsn/blob/d25fb320d4f0ef00168a597512a8b49588d0e304/crates/tlsn/src/zk_aes_ctr.rs#L200-L201
> 
> Then we do not have to check for which error variant it is, at higher levels.

^--- with the above approach we wouldn't be able to be certain whether the error comes from `zk_aes_ctr_sent` or `zk_aes_ctr_recv` (crates/tlsn/src/prover.rs), so the error message would have to be made more vague i.e. "check either `max_recv_data` or `max_sent_data`".
With my approach the error msg is more specific.

PR #1023, opened 2025-10-15
refactor(tlsn): invert control of config validation

### LINKED ISSUE #986: Invert control in verifier API
Right now we expose a `ProtocolConfigValidator` type, and also a `VerifyConfig` which is currently a placeholder.

Instead of having this logic injected into our API, we should yield control back out to the user. We can do this simply by adding an additional type state which exposes the config that needs validation by the user.

E.g.

```rust
impl Verifier<state::Validate> {
    fn config(&self) -> ProtocolConfig {  ..  }

    async fn accept(self) -> Result<Verifier<state::Online>, Error> { .. }

    async fn reject(self, msg: String) -> Result<(), Error> { .. }
}
```

sinui0:
Closes #986 

This PR refactors the verifier to move control of validation outside our API. This provides more flexibility at the expense of a bit of API complexity

This change is needed to support the SDK which needs to be able to delegate the validation logic across the ABI boundary.

### DIFF
--- a/crates/tlsn/src/msg.rs
+++ b/crates/tlsn/src/msg.rs
@@ -34,9 +34,7 @@ impl From<RejectionReason> for crate::prover::ProverError {
         if let Some(msg) = value.0 {
             crate::prover::ProverError::config(format!("verifier rejected with reason: {msg}"))
         } else {
-            crate::prover::ProverError::config(format!(
-                "verifier rejected without providing a reason"
-            ))
+            crate::prover::ProverError::config("verifier rejected without providing a reason")
         }
     }
 }


crates/wasm/src/verifier/mod.rs
```diff
@@ -49,8 +47,10 @@ impl State {
 impl JsVerifier {
     #[wasm_bindgen(constructor)]
     pub fn new(config: VerifierConfig) -> JsVerifier {
+        let tlsn_config = tlsn::verifier::VerifierConfig::builder().build().unwrap();
         JsVerifier {
-            state: State::Initialized(Verifier::new(config.into())),
+            state: State::Initialized(Verifier::new(tlsn_config)),
+            config,
```
th4s: this was confusing to me. why not expose the config like we do in tlsn?

th4s (approved):
:+1: lgtm. Some questions/remarks.

crates/wasm/src/verifier/mod.rs
sinui0: It's because we're going to delete `tlsn-wasm`. I only made these changes to ensure it continued to compile and behave the same.

PR #1027, opened 2025-10-20
Make tls_client compatible with a synchronous API

th4s:
This PR makes `write_plaintext` synchronous and removes some unneeded parts.

### DIFF
--- a/crates/tls/client/src/conn.rs
+++ b/crates/tls/client/src/conn.rs
@@ -525,10 +525,7 @@ impl ConnectionCommon {
 
     /// Writes the entire plaintext `buf` into an internal buffer.
     pub fn write_all_plaintext(&mut self, buf: &[u8]) -> Result<(), Error> {
-        let mut pos = 0;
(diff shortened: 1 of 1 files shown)


crates/tls/client/src/conn.rs
```diff
@@ -508,25 +511,25 @@ impl ConnectionCommon {
...
     }
 
-    /// Write entire buffer into connection.
-    pub async fn write_all_plaintext(&mut self, buf: &[u8]) -> Result<usize, Error> {
+    /// Writes the entire plaintext `buf` into an internal buffer.
+    pub fn write_all_plaintext(&mut self, buf: &[u8]) -> Result<(), Error> {
         let mut pos = 0;
         while pos < buf.len() {
-            pos += self.write_plaintext(&buf[pos..]).await?;
+            pos += self.write_plaintext(&buf[pos..])?;
```
sinui0: This will loop forever if the internal buffer does not grow to fit all the plaintext

PR #1031, opened 2025-10-21
feat(tlsn): insecure mode

### LINKED ISSUE #1029: Support "debug mode" for tests
Now that we have an "ideal" VM available for testing, we should provide a build configuration option which uses it in `tlsn`. This will allow us (and our users) to run integration tests without paying for the cryptography overhead.

sinui0:
Closes #1029 

This PR adds an "insecure" mode which can be enabled by compiling `tlsn` with the `tlsn_insecure` flag. This will compile using the `IdealVm` from `mpz` which avoids performing a bunch of expensive cryptography.

A follow up PR can also replace COT protocols with ideal replacements.

### DIFF
--- a/crates/tlsn/Cargo.toml
+++ b/crates/tlsn/Cargo.toml
@@ -40,6 +40,7 @@ mpz-ole = { workspace = true }
 mpz-ot = { workspace = true }
 mpz-vm-core = { workspace = true }
 mpz-zk = { workspace = true }
+mpz-ideal-vm = { workspace = true }
 
 aes = { workspace = true }
 ctr = { workspace = true }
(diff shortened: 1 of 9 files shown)


sinui0:
> what is the reason for using compile flags instead of feature flags here?

My rationale was the following:

Feature flags are additive across the entire dependency graph. Enabling such a feature by accident anywhere would be disastrous. A supply chain exploiter that obtains publish access to _any transitive dependency_ could add `tlsn` as a dependency and enable that feature causing all upstream builds to have it enabled. A compiler flag ensures that the decision must be made at compile time instead.

PR #1036, opened 2025-10-27
refactor: sans-io TLS IO

th4s:
This PR adds a sans-io api for our prover in the form of a new prover state `Prover<Connected>`.

### DIFF
--- a/crates/tlsn/src/prover/client/mpc.rs
+++ b/crates/tlsn/src/prover/client/mpc.rs
@@ -262,9 +262,12 @@ impl TlsClient for MpcTlsClient {
             State::Busy { mut mpc, mut fut } => {
                 trace!("inner client is busy");
 
-                // mpc future cannot be ready at this point becaus we have not called `stop`
-                // yet.
-                let _ = mpc.as_mut().poll(cx)?;
+                let mpc_poll = mpc.as_mut().poll(cx)?;
+
+                assert!(
+                    matches!(mpc_poll, Poll::Pending),
+                    "mpc future should not be finished here"
+                );
 
                 match fut.as_mut().poll(cx)? {
                     Poll::Ready(inner) => {


crates/tlsn/src/prover.rs
```diff
@@ -202,95 +198,135 @@ impl Prover<state::CommitAccepted> {
...
+    pub fn wants_write_tls(&self) -> bool {
+        self.state.tls_client.wants_write_tls()
+    }
 
-                let (_, (mut ctx, tls_transcript)) = futures::try_join!(
-                    conn_fut,
-                    mpc_fut.in_current_span().map_err(ProverError::from)
-                )?;
+    /// Reads TLS data from the server.
+    pub fn read_tls(&mut self, buf: &[u8]) -> Result<usize, std::io::Error> {
```
sinui0: Upon further thought, instead of using `std::io::Error` it probably makes more sense to use `ProverError`

crates/tlsn/src/prover.rs
```diff
@@ -202,95 +198,135 @@ impl Prover<state::CommitAccepted> {
...
+    ///
+    /// This is a performance optimization which will significantly reduce the
+    /// amount of upload bandwidth used by the prover.
+    ///
+    /// # Notes
+    ///
+    /// * The prover may need to close the connection to the server in order for
+    ///   it to close the connection on its end. If neither the prover or server
+    ///   close the connection this will cause a deadlock.
+    pub async fn defer_decryption(&mut self) -> Result<(), ProverError> {
```
sinui0: Let's change this to `enable_decryption(enable: false)`, and add an `is_decrypting` method.

Also, this method should not be async. So we'll need to move the handling of this into `poll`

crates/tlsn/src/prover.rs
```diff
@@ -202,95 +194,151 @@ impl Prover<state::CommitAccepted> {
...
+                Poll::Ready(Ok(()))
             }
-            .instrument(span)
-        });
+            Poll::Pending => Poll::Pending,
+        }
+    }
+
+    /// Returns a committed prover after the TLS session has completed.
+    pub fn finish(self) -> Option<Prover<state::Committed>> {
```
sinui0:
```suggestion
    pub fn finish(self) -> Result<Prover<state::Committed>, ProverError> {
```

so we can return the reason why the state transition failed

crates/tlsn/src/prover/client/mpc.rs
```diff
@@ -0,0 +1,470 @@
...
+    Busy {
+        fut: Pin<Box<dyn Future<Output = Result<Box<InnerClient>, ProverError>>>>,
+    },
+    Error,
+}
+
+struct InnerClient {
+    status: Status,
+    span: Span,
+    waker: Option<Waker>,
```
sinui0: I don't understand the purpose of this waker.

crates/tlsn/src/prover/client/mpc.rs
```diff
@@ -0,0 +1,474 @@
...
+                    fut: Box::pin(inner.run()),
+                };
+                self.poll(cx)
+            }
+            State::Busy { mut mpc, mut fut } => {
+                trace!("inner client is busy");
+
+                // mpc future cannot be ready at this point becaus we have not called `stop`
+                // yet.
+                let _ = mpc.as_mut().poll(cx)?;
```
sinui0: Assert this

Issue #622, opened 2024-10-03
Add signature verification during attestation validation


themighty1:
When validating the attestation, the Notary's signature verification is missing. We should add it to https://github.com/tlsnotary/tlsn/blob/b724d6a1d2705b682c4c71e1d81ce7bde92b055d/crates/core/src/request.rs#L47

PR #794, opened 2025-04-22
fix(core): add attestation sig validation

themighty1:
This PR adds the missing validation of attestation signature.

Closes #622

PR #1037, opened 2025-10-27
fix(attestation): verify sig during validation

themighty1:
This PR adds the missing verification of the signature during attestation validation.
I realized that we can use a simpler approach than what we discussed in https://github.com/tlsnotary/tlsn/pull/794/

Closes: https://github.com/tlsnotary/tlsn/issues/622


themighty1:
all feedback addressed, merging

PR #1040, opened 2025-10-30
feat: added plugin-core crate

themighty1:
This PR adds a crate with core types of the plugin.

I kept the “plugin” semantics in this crate, as it didn't seem that introducing a level of abstraction to make it independent of plugin-specific concepts was justified.


crates/data-fixtures/data/http/response_json
```diff
@@ -3,4 +3,4 @@ Cookie: very-secret-cookie
 Content-Length: 44
 Content-Type: application/json
 
-{"foo": "bar", "bazz": 123, "buzz": [1,"5"]}
+{"foo": "bar", "bazz": 123, "buzz": [1,"5"]}
```
themighty1: I removed this newline on purpose, otherwise content length was incorrect.

Issue #959, opened 2025-08-12
multiple runs of the wasm build cause panic


themighty1:
### Problem

`wasm-bindgen` has a limitation - it does not call thread-local destructors. This causes the wasm build of `tlsn` to panic after a few runs.

### Details

Acc.to this PR's https://github.com/wasm-bindgen/wasm-bindgen/pull/2769 comment, `wasm-bindgen` does not have the ability to call TLS destructors:

"This doesn't run any destructors or dispose of TLS storage, which seems like a significantly more difficult problem"

The `tracing` crate which we rely on uses thread locals heavily. Due to the TLS destructors not being called,  `sharded_slab` (a `tracing` dependency) will panic with
> Thread count overflowed the configured max count. Thread index = 128, max threads = 128.
in https://github.com/hawkw/sharded-slab/blob/e540cdb7daafd6a6e9d17408d3de017029a6637f/src/shard.rs#L297

Full stacktrace:
<details>
```
$__rustc[a5b89e8e8011757d]::__rust_start_panic	@	harness_executor_bg.wasm:0x61c4a4
$__rustc[a5b89e8e8011757d]::rust_panic	@	harness_executor_bg.wasm:0x604700
$std::panicking::rust_panic_with_hook::hd0da58a2a1a3ca63	@	harness_executor_bg.wasm:0x4ee5df
$std::panicking::begin_panic_handler::{{closure}}::h2e2703491d4087bb	@	harness_executor_bg.wasm:0x584d5e
$std::sys::backtrace::__rust_end_short_backtrace::h7b5d1a1a58dbd50f	@	harness_executor_bg.wasm:0x61bfc6
$__rustc[a5b89e8e8011757d]::rust_begin_unwind	@	harness_executor_bg.wasm:0x601f50
$core::panicking::panic_fmt::h2899aa4559e5a097	@	harness_executor_bg.wasm:0x601fd9
$sharded_slab::pool::Pool<T,C>::create::hf61528a7b341bc94	@	harness_executor_bg.wasm:0x445eb1
$sharded_slab::pool::Pool<T,C>::create_with::hf3cef710a37063ea	@	harness_executor_bg.wasm:0x5596c8
$<tracing_subscriber::registry::sharded::Registry as tracing_core::subscriber::Subscriber>::new_span::h3f64b167989ea65c	@	harness_executor_bg.wasm:0x570a49
[...]

PR #1050, opened 2025-11-25
fix(wasm): use patched sharded-slab with manual thread state release

themighty1:
This PR patches a transitive dep `sharded-slab` in order to allow our wasm build to manually call destructors of objects in  thread-local storage.
Fixes: https://github.com/tlsnotary/tlsn/issues/959

Depends on https://github.com/tlsnotary/tlsn-utils/pull/73

Tested by running the wasm prover 100+ times (whereas before it would panic on the 8th run).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6640,8 +6640,7 @@ dependencies = [
 [[package]]
 name = "sharded-slab"
 version = "0.1.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f40ca3c46823713e0d4209592e8d6e826aa57e928f09752619fc696c499637f6"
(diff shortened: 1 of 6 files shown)


themighty1:
To follow up on the mattermost discussion. The actual patch is located here https://github.com/tlsnotary/sharded-slab/blob/5dd1d23d5b1f57e52b1858cb3c42ef4748062f3b/src/tid.rs#L18

What this patch does NOT do:
- it does not free memory, so there is no concern about freed memory being referenced later.

What this patch DOES:
- release the thread ID back into the pool, otherwise the pool will be starved and we will get a panic.

I additionally checked just in case that:
- a thread never looks up another thread's ID. Only the actual thread itself looks up its own ID.
- if the thread ID gets released and after that (highly unlikely) the thread looks up its ID, it will get a fresh ID.

PR #1052, opened 2025-11-26
clippy: allow deprecated aead::generic_array

heeckhau:


th4s:
I think the "right" fix would be to upgrade our `generic-array` dependency, or is there a reason this is not possible?

PR #1067, opened 2025-12-15
Feat/socket api

th4s:
This PR makes our API work with socket references instead of consuming them.


th4s:
@sinui0 you can take a look. We still need to patch dependencies correctly and open a PR for `better-futures-plex`.

Issue #1038, opened 2025-10-27
Refactor protocol IO


sinui0:
Currently our API takes ownership over the IO handle between the Prover and Verifier. Instead, our API should provide a poll-oriented buffered IO interface which does not require ownership of the platform IO handle.

sinui0:
We decided to not go with `sans-io`, but instead we do allow the IO to be reclaimed from the `Session` type.

PR #1075, opened 2026-01-14
feat(tlsn): session type

### LINKED ISSUE #1039: Distinguish between user error and internal errors
The user should be able to distinguish between errors arising from user input, or from internal errors (e.g. bugs). This also includes being able to distinguish when the verifier rejects some proposed configuration.

sinui0:
This PR provides a few things:

1. Introduces the `Session` type which manages session-level state such as the IO between the prover and verifier. In the future this can also contain state such as the VOLE secret sharings which can be re-used between instances. This type has the feature of enabling the IO to be reclaimed after the user is done with our API closing #1038 
2. Updates the multiplexer dependency
3. Consolidates errors into a single crate-level type and also closes #1039

### DIFF
--- a/crates/examples/interactive_zk/prover.rs
+++ b/crates/examples/interactive_zk/prover.rs
@@ -41,8 +41,8 @@ use tlsn::{
     Session,
 };
 
-use tlsn_examples::{MAX_RECV_DATA, MAX_SENT_DATA};
 use futures::io::AsyncWriteExt as _;
+use tlsn_examples::{MAX_RECV_DATA, MAX_SENT_DATA};
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_util::compat::{FuturesAsyncReadCompatExt, TokioAsyncReadCompatExt};
 use tracing::instrument;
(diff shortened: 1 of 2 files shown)


crates/examples/interactive_zk/prover.rs
th4s: Would be better to remove the `verifier_extra_socket` to show that returning the socket and using it for other stuff works.

crates/tlsn/src/session.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /// Creates a new prover.
+    pub fn new_prover(
+        &mut self,
+        config: ProverConfig,
+    ) -> Result<Prover<prover_state::Initialized>> {
+        let ctx = self.mt.new_context().map_err(|e| {
+            Error::internal()
+                .with_msg("failed to created new prover")
```
th4s:
```suggestion
                .with_msg("failed to create new prover")
```

crates/tlsn/src/session.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /// Creates a new verifier.
+    pub fn new_verifier(
+        &mut self,
+        config: VerifierConfig,
+    ) -> Result<Verifier<verifier_state::Initialized>> {
+        let ctx = self.mt.new_context().map_err(|e| {
+            Error::internal()
+                .with_msg("failed to created new verifier")
```
th4s:
```suggestion
                .with_msg("failed to create new verifier")
```

PR #1084, opened 2026-01-22
fix(harness): fix browser latency emulation and enable TCP nodelay

themighty1:
This PR fixes a bug in the harness due to which the delay was not applied in the Prover->Verifier direction in browser benches.

Before this PR, when RTT was expected to be 400ms :
```
Browser path (only verifier→websocket-relay has delay):                                                                                                   
websocket-relay(host) → bridge → veth_proto_1.1 → veth_proto_1.0 → verifier(ns_1)     [NO delay]                                                                           
verifier(ns_1) → veth_proto_1.0[+200ms] →  veth_proto_1.1 → bridge → websocket-relay(host)  [+200ms]                                                                        
RTT = ~200ms (asymmetric) 
```

The solution in this PR is to run websocket-relay inside a namespace:
```
websocket-relay(ns_0) → veth_proto_0.0[+200ms] → bridge → veth_proto_1.1 → veth_proto_1.0 → verifier(ns_1) [+200ms]                                                                                         
verifier(ns_1) → veth_proto_1.0[+200ms] → veth_proto_1.1 → bridge → veth_proto_0.0 → websocket-relay(ns_0) [+200ms]
RTT = ~400ms            
 ```

Also removes Nagel from the app server IO while we are at it.

I analyzed a network trace before and after this PR to double-check that the diagrams above are correct.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8602,7 +8602,7 @@ dependencies = [
 [[package]]
 name = "websocket-relay"
 version = "0.1.0"
-source = "git+https://github.com/tlsnotary/tlsn-utils?rev=6f1a934#6f1a934bc4ce358c268b26abf4fd086267b91c70"
+source = "git+https://github.com/tlsnotary/tlsn-utils?rev=92197f9#92197f928b08cd3a1c799d513347c56881db1cb6"
 dependencies = [
  "anyhow",
  "form_urlencoded",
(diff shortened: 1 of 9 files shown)


themighty1:
This was a mis-diagnosis, the correct fix is at https://github.com/tlsnotary/tlsn/pull/1091

PR #1100, opened 2026-02-10
refactor: MPC-TLS actor

th4s:
This PR fixes the hanging browser benches by removing the actor in `mpc-tls/leader`.

The problem is that in the browser's event loop, futures are scheduled as microtasks. Each microtask is able to schedule another microtask (directly being executed) and if repeated, this can lead to an infinite loop blocking macrotasks (e.g. socket reads) and other microtasks. This is not true for native, where `Pending` means yielding to the executor even when a task reschedules itself.

Here the `tls-client` `poll` was communicating with the `mpc-tls-leader` by the indirection of the actor, which itself uses `async/await` cooperative task scheduling. This lead to an infinite micro-task queue, because the actor and the `tls-client` are sitting on the same task, which blocked the multiplexer from making progress.

This PR:
- removes our actor framework `ludi` entirely (was planned anyway and was now causing problems)
- adapts the `TlsClient` implementation  to this change (no more mpc-polling necessary)
- removes the `Backend` crate and migrates the `Backend` trait into `tls/client`.

Benching native and browser shows no impact on runtime performance.

### DIFF
--- a/crates/tlsn/src/prover/client/mpc.rs
+++ b/crates/tlsn/src/prover/client/mpc.rs
@@ -341,19 +341,32 @@ impl InnerState {
 
     #[instrument(parent = &self.span, level = "trace", skip_all, err)]
     async fn run(mut self: Box<Self>) -> Result<Box<Self>, TlsnError> {
-        self.tls
+        let mut state = self
+            .tls
             .process_new_packets()
             .await
             .map_err(|err| TlsnError::internal().with_source(err))?;
+        loop {
+            let new_state = self
+                .tls
+                .process_new_packets()
+                .await
+                .map_err(|err| TlsnError::internal().with_source(err))?;
+
+            if new_state.plaintext_bytes_to_read() == state.plaintext_bytes_to_read()
+                && new_state.tls_bytes_to_write() == state.tls_bytes_to_write()
+            {
+                break;
+            }
+            state = new_state;
+        }
+
         Ok(self)
     }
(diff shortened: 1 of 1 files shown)


th4s:
Alright then I will wait :+1:

I took this as a motivation to re-investigate again this morning. I co-debugged this with my clanker and I think the following makes sense to me:

> The new prover architecture has an infinite microtask loop in WASM. Here's why:

  > MpcTlsClient always calls process_new_packets when polled in the Active state. process_new_packets communicates with the MPC future via async channels. Both the MPC future and the
  process_new_packets future are polled with the same cx.waker(). So any channel send immediately wakes the prover's own task, scheduling a new microtask.

 >  When process_new_packets completes, the state returns to Active. The pending self-wake fires, re-polls the prover, which re-polls tls_client.poll(), which sees Active, calls
  process_new_packets again — generating the next self-wake. This repeats forever.

>  In the old code, process_new_packets was gated behind select_biased!, which only called it in response to external events (server data, user data, notifications). A spurious self-wake would
  re-enter select_biased!, find no external events, and return Pending without calling process_new_packets. The chain died.

 >  The fix needs to either gate process_new_packets behind actual new data, or prevent the MPC channel communication from waking cx.waker().

This PR goes with option 2. Do you see an advantage in trying out option 1 instead? My thoughts were that this has to be done anyways at some point and it resolves some tech debt at the same time.

PR #1101, opened 2026-02-10
feat(harness): forward logging of prover's wasm ouput to CLI when benching browser target

th4s:
This PR adds forwarding wasm log from the browser to CLI when running the browser benches. This is only supported when building the harness with `./build.sh debug`.

It turned out to be useful when debugging the browser benches for deadlocks.


themighty1 (approved):
nice, thanks

PR #1122, opened 2026-03-10
feat: Proxy-TLS commitment protocol

th4s:
This PR adds the proxy-based approach to TLSNotary.

  - Core config (`crates/core`): Extends `TlsCommitProtocolConfig` with a Proxy variant and adds `ProxyTlsConfig`
  - Dependency wiring (`crates/tlsn/src/deps/`): Replaces the old monolithic `mpz.rs` with a `deps/` module that has `ProverDeps` and `VerifierDeps` enums, each with Mpc and Proxy variants,  encapsulating setup/allocation logic per mode.
  - Proxy TLS client (`crates/tlsn/src/prover/client/proxy/`): A new `ProxyTlsClient` implementing the `TlsClient` trait. Uses custom rustls `CryptoProvider` wrappers (`InterceptingKxGroup`, `InterceptingPrf`) to intercept the pre-master secret, verify-data, and handshake hashes during a real TLS 1.2 handshake.
  - Proxy protocol logic (`crates/tlsn/src/proxy/`): Core proxy module with `ProxyProver` and `ProxyVerifier` (ZK-based PRF verification and key derivation), plus a `TlsParser` that parses raw TLS records to reconstruct the `TlsTranscript`.
  - Prover/Verifier refactoring: Both Prover and Verifier now branch on the protocol config to instantiate either MPC or proxy dependencies. Tag verification is moved out of the MPC client into the shared finish() path. `ProverFuture` gains a Finishing state to handle the now-async finalization.
  - Verifier proxy forwarding (`crates/tlsn/src/verifier.rs`): In proxy mode, the Verifier connects to the server, forwards TLS traffic bidirectionally, parses the TLS records, and then runs ZK verification of the verify_data.
  - Tests, examples, harness, WASM: Adds a proxy integration test, a proxy example, harness benchmark/test plugin support, and wasm bindings for proxy mode configuration.


th4s:
## Benches

Config used: `bench.toml` (or `bench_proxy.toml`)

```
[[group]]
name = "cable"
bandwidth = 20
protocol_latency = 20
upload-size = 1024
download-size = 2048

[[bench]]
group = "cable"


[[group]]
name = "mobile_5g"
bandwidth = 30
protocol_latency = 30
upload-size = 1024
download-size = 2048

[[bench]]
group = "mobile_5g"


[[group]]
name = "fiber"
bandwidth = 100
protocol_latency = 15
upload-size = 1024
download-size = 2048

[[bench]]
group = "fiber"
```

## Results

### MPC Native
cable [mpc] (20 Mbps, 20ms latency, 1KB↑ 2KB↓):
  Median:  14.51s

fiber [mpc] (100 Mbps, 15ms latency, 1KB↑ 2KB↓):
  Median:  3.65s

mobile_5g [mpc] (30 Mbps, 30ms latency, 1KB↑ 2KB↓):
  Median:  10.38s

### MPC Browser
cable [mpc] (20 Mbps, 20ms latency, 1KB↑ 2KB↓):
  Median:  15.79s

fiber [mpc] (100 Mbps, 15ms latency, 1KB↑ 2KB↓):
  Median:  4.98s

mobile_5g [mpc] (30 Mbps, 30ms latency, 1KB↑ 2KB↓):
  Median:  11.64s

### Proxy Native
cable_proxy [proxy] (20 Mbps, 20ms latency, 1KB↑ 2KB↓):
  Median:  1.57s

fiber_proxy [proxy] (100 Mbps, 15ms latency, 1KB↑ 2KB↓):
  Median:  0.94s

mobile_5g_proxy [proxy] (30 Mbps, 30ms latency, 1KB↑ 2KB↓):
  Median:  1.55s

### Proxy Browser
cable_proxy [proxy] (20 Mbps, 20ms latency, 1KB↑ 2KB↓):
  Median:  2.04s

fiber_proxy [proxy] (100 Mbps, 15ms latency, 1KB↑ 2KB↓):
  Median:  1.40s

mobile_5g_proxy [proxy] (30 Mbps, 30ms latency, 1KB↑ 2KB↓):
  Median:  1.95s

th4s:
### TODOs before merge
* [x] Review and merge https://github.com/privacy-ethereum/mpz/pull/378 and adapt `mpz` dependencies in `Cargo.toml` to correct tag/rev.
* [x] Review and merge https://github.com/tlsnotary/tlsn/pull/1121 and rebase.

crates/tlsn/tests/proxy_test.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    tls_connection
+        .write_all(b"GET / HTTP/1.1\r\nConnection: close\r\n\r\n")
+        .await
+        .unwrap();
+
+    let mut response = vec![0u8; 1024];
+    tls_connection.read_to_end(&mut response).await.unwrap();
+
+    tls_connection.close().await.unwrap();
```
th4s: Ideally we want to `write`, `close` and then `read`. I did not find the right solution to do this, so left it at `write`, `read` and `close` for now.

The problem: Even though packets arrive in the correct order on the server side (I checked) namely first `app_data` then `close_notify_alert`, the server parses it in one go and because he sees the close_notify he does **NOT** answer the app_data request.

The problem is not visible on the MPC test I guess, because there is an async latency overhead because of the network topology. The server parses the records one after the other and thus sends a response.

Not clear what the best approach here is? Introducing artificial time-based delays seems wrong and would not be a proper solution. The right fix is on the server side I think, but the question  is how do we want to handle this on the client side?

crates/tlsn/tests/proxy_test.rs
sinui0, line 1: We shouldn't need a duplicate integration test file for this, most of the test code is the same

Cargo.toml
```diff
@@ -140,18 +140,21 @@ regex = { version = "1.10" }
...
+spki = { version = "0.8", features = ["std"] }
+const-oid = { version = "0.10", features = ["db"] }
 semver = { version = "1.0" }
 serde = { version = "1.0" }
 serde_json = { version = "1.0" }
 sha2 = { version = "0.10" }
 signature = { version = "2.2" }
 thiserror = { version = "1.0" }
 tiny-keccak = { version = "2.0" }
+tls-parser = { version = "0.12" }
```
sinui0, line 157: We already have TLS parsing in tls-core

crates/core/src/config/tls_commit/proxy.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use crate::{config::tls_commit::NetworkSetting, connection::DnsName};
+use serde::{Deserialize, Serialize};
+
+/// Proxy-TLS commitment protocol configuration.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ProxyTlsConfig {
+    /// Whether the `deferred decryption` feature is toggled on from the start
+    /// of the TLS connection.
+    defer_decryption_from_start: bool,
```
sinui0: This option does not make sense in the proxy setting. The Prover knows the encryption key and can decrypt freely.

crates/core/src/config/tls_commit/proxy.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Network settings.
+    network: NetworkSetting,
```
sinui0: Similarly, this is an MPC commitment protocol setting. Not needed

crates/core/src/config/verifier.rs
```diff
@@ -35,12 +45,74 @@ impl VerifierConfigBuilder {
...
+#[derive(Default, Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub enum ConnectionMode {
+    /// Only accepts multi-party computation.
+    #[default]
+    Mpc,
+    /// Only accepts proxy mode.
+    Proxy,
+    /// Accepts both modes.
+    Universal,
+}
```
sinui0: This can be eliminated with the proposed API changes, which moves direct control back out to the user.

crates/core/src/lib.rs
```diff
@@ -15,13 +15,30 @@ pub use rangeset;
 pub mod config;
 pub(crate) mod display;
 
+pub use mpz_memory_core::{binary::U8, Array};
```
sinui0: No `mpz` dependency allowed in this crate.

crates/core/Cargo.toml
```diff
@@ -25,7 +25,9 @@ fixtures = [
 [dependencies]
 tlsn-data-fixtures = { workspace = true, optional = true }
 tlsn-tls-core = { workspace = true, features = ["serde"] }
+
 rangeset = { workspace = true, features = ["serde"] }
+mpz-memory-core = { workspace = true }
```
sinui0: 🙅

crates/harness/executor/src/bench/prover.rs
sinui0, line 1: Same as integration test comment, this should not require a new bench program just to switch the commitment protocol.

crates/harness/executor/test_plugins/proxy.rs
sinui0, line 1: ditto

crates/harness/bench_proxy.toml
sinui0, line 1: Commitment protocol is a config flag, does not need new set of bench definitions

crates/tlsn/src/prover/client/proxy/kx.rs
```diff
@@ -0,0 +1,148 @@
...
+            .kx_groups
+            .iter()
+            .filter(|g| groups.contains(&g.name()))
+            .map(|&inner| {
+                // Leaked to satisfy rustls's `&'static dyn SupportedKxGroup`
+                // requirement. ~64 bytes per group per session.
+                &*Box::leak(Box::new(InterceptingKxGroup {
+                    inner,
+                    pms: pms.clone(),
+                })) as &'static dyn SupportedKxGroup
```
sinui0: There must be a better way which does not introduce a memory leak.

crates/tlsn/src/prover/client/proxy/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+const ALLOWED_SUITES: &[CipherSuite] = &[
+    CipherSuite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+    CipherSuite::TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+];
+
+pub(crate) struct ProxyTlsClient {
+    conn: ClientConnection,
+    parser: TlsParser,
+    decrypt: Arc<DecryptState>,
+    decrypt_mark: usize,
```
sinui0: should not be needed

crates/tlsn/src/prover/client/proxy/prf.rs
```diff
@@ -0,0 +1,200 @@
...
+use std::sync::{Arc, OnceLock};
+
+#[derive(Default, Clone)]
+pub(crate) struct PrfTranscript {
+    client_finished: Arc<OnceLock<Vec<u8>>>,
+    server_finished: Arc<OnceLock<Vec<u8>>>,
+    client_handshake_hash: Arc<OnceLock<Vec<u8>>>,
+    server_handshake_hash: Arc<OnceLock<Vec<u8>>>,
+    session_hash: Arc<OnceLock<Vec<u8>>>,
+}
```
sinui0: All of this is derivable from the TLS transcript, we should not need to intercept it like this.

I think the only think we need to pull out of the TLS client is the PMS.

crates/tlsn/src/prover/future.rs
```diff
@@ -10,21 +10,31 @@ use crate::{
...
-
-        ProverControl { decrypt_state }
+    pub fn control(&self) -> Option<ProverControl> {
+        match &self.state {
+            FutureState::Connected { prover } => {
+                let decrypt_state = prover.state.tls_client.decrypt();
+                Some(ProverControl { decrypt_state })
+            }
+            _ => None,
+        }
```
sinui0: Why? Seems like a hack to get around an ownership/state modeling issue

crates/tlsn/src/proxy/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+            TlsnError::io()
+                .with_msg("receive cf_hash from prover failed")
+                .with_source(e)
+        })?;
+        let sf_hash: Vec<u8> = self.ctx.io_mut().expect_next().await.map_err(|e| {
+            TlsnError::io()
+                .with_msg("receive sf_hash from prover failed")
+                .with_source(e)
+        })?;
+        tracing::debug!("received handshake hashes");
```
sinui0: This does not seem correct. The Verifier holds the entire public TLS transcript and should not accept any values from the Prover.

The Prover's constraint is to commit to their private PMS and prove that they can produce an identical TLS transcript.

crates/tlsn/src/error.rs
```diff
@@ -119,6 +119,12 @@ impl From<std::io::Error> for Error {
     }
 }
 
+impl From<rustls::Error> for Error {
+    fn from(e: rustls::Error) -> Self {
+        Self::internal().with_msg("rustls error").with_source(e)
+    }
+}
```
sinui0: This expands the public API. Conversion should be internal. (We should get rid of all the `From` impls for `Error` at some point.)

crates/tlsn/src/verifier/state.rs
```diff
@@ -31,13 +28,18 @@ opaque_debug::implement!(CommitStart);
...
+    pub(crate) prover_socket: Option<BoxedIo>,
+    pub(crate) server_socket: Option<BoxedIo>,
 }
 
 opaque_debug::implement!(CommitAccepted);
 
+pub(crate) trait AsyncReadWrite: AsyncRead + AsyncWrite {}
+impl<T: AsyncRead + AsyncWrite + ?Sized> AsyncReadWrite for T {}
+
+type BoxedIo = Box<dyn AsyncReadWrite + Send + Unpin + 'static>;
```
sinui0: 🙅 we can avoid boxing with the suggested API changes

sinui0 (changes requested):
This approach requires the Prover and Verifier to establish a new connection with each other for every TLS commitment. Instead, we should utilize the existing connection (open a new mux stream) for the proxy.

Perhaps we should fork the Prover/Verifier APIs using type states, something like this:

```rust
trait CommitConfig {
  type Protocol;
}
struct Mpc;
struct Proxy;
impl CommitConfig for MpcTlsCommitConfig {
  type Protocol = Mpc;
}

impl Prover<state::Initialized> {
  pub async fn commit<C: CommitConfig>(self, config: C) -> Result<Prover<state::CommitAccepted<C::Protocol>>> {}
}

enum VerifierCommitRequest {
  Mpc(Verifier<state::CommitStart<Mpc>>),
  Proxy(Verifier<state::CommitStart<Proxy>>),
}

impl Verifier<state::Initialized> {
  pub async fn commit(self) -> Result<VerifierCommitRequest> {}
}

impl Verifier<state::CommitStart<Proxy>> {
  pub fn request(&self) -> ProxyConfig {}

  pub fn accept(self) -> Result<Verifier<state::CommitAccepted<Proxy>>> {}
}
```

What do you think?

---

crates/tlsn/src/proxy/verifier.rs
th4s: The issue is that to get `sf_hash` you need to decrypt encrypted parts of the transcript , and the verifier does not have the keys to decrypt. We deal similarly with `cf_hash` and `sf_hash` in the MPC protocol?

crates/tlsn/src/prover/client/proxy/prf.rs
th4s: yeah it is simpler to do it like this, because you let rustls do the decryption and hashing work.

crates/core/src/config/tls_commit/proxy.rs
th4s: Hmm the `NetworkSetting` has an effect on the PRF circuit size, which is also used in proxy mode. Do you think that this effect is negligible considering performance, since we are not dealing with garbling? What should be used then for proxy mode? Full PRF, bigger circuit, less roundtrips?

sinui0: Just prove the full PRF, it is negligible

crates/tlsn/src/prover/client/proxy/prf.rs
sinui0: We need it implemented for the Verifier anyhow (e.g. hashing)

th4s: I reduced what is intercepted to just `client_handshake_hash` and `server_handshake_hash`. Especially `server_handshake_hash` cannot be reconstructed by the verifier from the transcript.

th4s: removed it completely

crates/tlsn/src/proxy/verifier.rs
th4s: Works now, just PMS as private input.

crates/core/src/transcript/tls.rs
sinui0, line 1: Need to put more thought into the public API here. TlsTranscript has some awkward associated methods such as TlsTranscript::parse_handshake() which would make more sense as a standalone function (if it were something we actually wanted to expose, see below).
It seems to me we should only need TlsTranscript::parse and then getters for data such as cf_hash (e.g `TlsTranscript::cf_hash(&self) -> Option<_>`.

crates/core/src/transcript/tls.rs
```diff
@@ -360,12 +610,646 @@ impl TlsTranscriptError {
...
+    use crate::connection::KeyType;
+    use rustls::ClientConnection;
+    use std::{
+        io::{Read, Write},
+        sync::Arc,
+    };
+    use tls_server_fixture::{CA_CERT_DER, SERVER_CERT_DER, SERVER_DOMAIN};
+    use tokio_util::{compat::TokioAsyncReadCompatExt, io::SyncIoBridge};
+
+    #[tokio::test]
```
sinui0: We should not be pulling in `tokio` to run these kinds of unit tests. Capture some fixtures instead

Cargo.toml
sinui0: stale?

crates/tlsn/src/prover/client/proxy/kx.rs
```diff
@@ -0,0 +1,148 @@
...
+//! `PMS` and `GROUPS` are process-wide singletons. This is a consequence of
+//! rustls requiring `&'static dyn SupportedKxGroup` references in its
+//! `CryptoProvider`, which prevents per-session allocation.
+//!
+//! **Only one proxy-mode TLS session may be active per process at a time.**
+//! Running concurrent proxy sessions will cause one session to observe the
+//! other's pre-master secret (or an empty vec), leading to protocol failure.
+//! `GROUPS` is initialized once via `OnceLock` and reused for all subsequent
+//! sessions — the first `CryptoProvider` configuration wins for the process
+//! lifetime.
```
sinui0: This is pretty clearly fragile and not ok

[`This feature`](https://docs.rs/rustls/latest/rustls/trait.KeyLog.html) is likely a better option for what we want to do. The PMS is never used to compute anything other than the master secret, so we should be able to skip that PRF stage (prover provides the MS as private input). TLS 1.3 should also be possible using this trait.

crates/tlsn/src/proxy/prover.rs
```diff
@@ -0,0 +1,209 @@
...
+        )?;
+
+        tracing::info!("Proxy TLS done");
+        let output = TlsOutput {
+            keys: refs.keys,
+            tls_transcript,
+        };
+
+        Ok((self.ctx, self.vm, output))
+    }
```
sinui0, line 176: All inputs/outputs are known prior to calling this, we should not need to flush multiple times.

crates/tlsn/src/prover.rs
```diff
@@ -247,24 +266,23 @@ impl Prover<state::CommitAccepted> {
...
 }
 
-impl<S> Future for Prover<state::Connected<S>>
+impl<S> Prover<state::Connected<S>>
 where
-    S: AsyncRead + AsyncWrite + Send + Unpin + 'static,
+    S: AsyncRead + AsyncWrite + Send + Unpin,
 {
-    type Output = Result<(), Error>;
+    /// Returns  a [`ProverControl`] for connection specific settings.
```
sinui0:
```suggestion
    /// Returns a [`ProverControl`] for connection specific settings.
```

crates/tlsn/src/prover.rs
```diff
@@ -247,24 +266,23 @@ impl Prover<state::CommitAccepted> {
...
+    ///
+    /// Only provided for MPC mode.
```
sinui0: Why?

sinui0 (changes requested):
What did you determine while trying to use typestates to model the API differences between MPC vs proxy?

crates/tlsn/src/proxy/prover.rs
th4s: `cf_vd` is required as an input to `TlsTranscript::compute_sf_hash`. The verifier is not able to compute `sf_hash` without it, since `cf_vd` is encrypted on the wire.

crates/tlsn/src/prover.rs
th4s: Is the goal here to provide a `ProverControl` also in proxy mode? But then it does nothing?

th4s:
> What did you determine while trying to use typestates to model the API differences between MPC vs proxy?

The main problem is that you also need to pass the connection/socket and this is different depending on the side and mode

Prover: In MPC mode needs a socket and in proxy mode not. The socket type also leaks into the prover type.
Verifier: In proxy mode needs  a socket and in MPC mode not. No type leakage since it is not stateful.

Typestates felt like trying to work around something, and the end result did not provide any real benefit plus extra code and abstraction. If we decide to do it with typestates this propagates into the whole api. One advantage of the existing approach is, that it is very explicit. You can basically grep for what protocol is used.

Cargo.toml
th4s: yep

crates/tlsn/src/prover/client/proxy/kx.rs
th4s: In this case we no longer prove the derivation of the master_secret. That should be ok, but it feels weird to relax our security guarantees to be able to use upstream rustls. This will also require changes to our PRF.

crates/core/src/transcript/tls.rs
th4s: The problem is that we need the verify_data **before** being able to call `TlsTranscript::parse`.

crates/tlsn/src/prover/client/proxy/kx.rs
th4s: Hmm I went through all the options we have again. It seems that providing the MS seems to be the best alternative. rustls api sucks.

crates/tlsn/src/prover.rs
sinui0: Yes, it may be extended in the future for actions that apply to both. Having `Option` depending on what mode was selected elsewhere is more awkward than the decryption control being a noop.

crates/core/src/config/tls_commit.rs
```diff
@@ -1,52 +1,51 @@
...
 
 use serde::{Deserialize, Serialize};
 
+use crate::config::tls_commit::{mpc::MpcTlsConfig, proxy::ProxyTlsConfig};
+
 /// TLS commitment configuration.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct TlsCommitConfig {
-    protocol: TlsCommitProtocolConfig,
+pub struct TlsCommitConfig<P> {
```
sinui0: What purpose does this type serve?

crates/tlsn/src/deps.rs
```diff
@@ -0,0 +1,70 @@
...
+use tlsn_deap::Deap;
+
+mod prover;
+pub(crate) use prover::{ProverMpc, ProverMpcDeps, ProverProxyDeps, ProverZk};
+
+mod verifier;
+pub(crate) use verifier::{VerifierMpcDeps, VerifierProxyDeps, VerifierZk};
+
+/// Trait for protocol dependencies.
+pub trait ProtocolDeps {
```
sinui0: This abstraction seems unnecessary, why do we need a trait when there are 2 implementations? Leaking this into the public API is also not great

crates/tlsn/src/verifier.rs
```diff
@@ -110,50 +120,32 @@ impl Verifier<state::CommitStart> {
 
     /// Accepts the proposed protocol configuration.
     #[instrument(parent = &self.span, level = "info", skip_all, err)]
-    pub async fn accept(mut self) -> Result<Verifier<state::CommitAccepted>> {
+    pub async fn accept<P: ProtocolConfig>(
```
sinui0: This requires the Verifier to clone and pass the config back in to itself, clunky (and allows a confused user to pass construct their own new config and pass it in).

What was wrong with my suggestion of `Verifier::commit -> enum { Mpc(Verifier<state::CommitStart<Mpc>), Proxy(Verifier<state::CommitStart<Proxy>) }`?

crates/core/src/transcript/tls.rs
sinui0: Can you elaborate?

TlsTranscript should be constructed after the entire transcript is recorded/committed. It shouldn't be used as a stateful data structure elsewhere

sinui0 (changes requested):
My concerns are mostly still just on the public API of the Prover/Verifier types in `tlsn` and the `TlsTranscript` type in `tlsn-core`.

The Prover/Verifier API looks close to me, the API fork with typestates seems clean enough. Just need to tighten it a bit.

Using `TlsTranscript` in `tlsn` as a stateful data structure for computing the transcript hashes seems to be the culprit for why its API blew up. Should be able to simplify that

crates/tlsn/src/verifier.rs
th4s: Agree, I think what drove me here was symmetry to the prover api. Let me fix this and give up the symmetry.

crates/core/src/config/tls_commit.rs
th4s: are you talking about `TlsCommitConfig`?, yeah we could get rid of this type,  I think.

crates/tlsn/src/deps.rs
th4s: The question is how to fork the prover, or how should `Prover::commit` look like. Currently it is generic over `TlsCommitConfig<P: ProtocolConfig>`. Like you mentioned above we could just use `P` and get rid of `TlsCommitConfig`, but irrespective of that the main challenge is:

How to "connect" the public config types `MpcTlsConfig`/`ProxyTlsConfig` with their dependencies, which we want to `setup` internally, without leaking into the public api.

crates/core/src/transcript/tls.rs
th4s: I think the root cause is that `TlsTrascript::parse` internally calls `TlsTranscript::new` to make use of some correctness checks. The problem is that `TlsTranscript::new` requires cleartext verify data to perform some internal checks, and this is not available for the verifier.

There are some ways to work around this. What do you think? For example we could make the vd_check in `TlsTanscript::new` optional, so that it accepts a `verify_data: Option<VerifyData>`?

crates/tlsn/src/deps.rs
th4s: I proposed an improved API in the most recent commits.

crates/core/src/config/tls_commit.rs
sinui0: yes, we can get rid of it. It is just a newtype around a single generic field at this point

crates/tlsn/src/verifier.rs
```diff
@@ -97,10 +106,18 @@ impl Verifier<state::Initialized> {
...
         })
     }
 }
+/// Commit accepted verifiers for different protocols.
+pub enum VerifierCommitAccepted {
+    /// Verifier for MPC protocol.
+    Mpc(Verifier<state::CommitAccepted<MpcTlsConfig>>),
+    /// Verifier for Proxy protocol.
+    Proxy(Verifier<state::CommitAccepted<ProxyTlsConfig>>),
+}
```
sinui0: This requires the user to match on whether it is proxy vs MPC twice. The enum should wrap `Verifier<state::CommitStart<P>>`

th4s: I don't think, you need to match twice: https://github.com/tlsnotary/tlsn/blob/feat/proxy-approach/crates/tlsn/tests/utils.rs#L130-L153

sinui0: this example does not have the verifier check the config (which needs a match), it unconditionally accepts

PR #1127, opened 2026-04-14
feat(sdk-core,wasm): add hash-commitment support for handlers

heeckhau:
## Summary

- Wire the `prove_hash()` path through the SDK layer so plugins can use `action: "HASH"` on handlers to hash-commit transcript ranges instead of revealing them as plaintext
- Add `HashAlgorithm` enum (Blake3/Sha256/Keccak256) — required on HASH handlers, no default
- `CommitRange` uses `Range<usize>` with non-optional `HashAlgorithm`
- `HandlerAction` is a data-carrying enum: `Reveal` or `Hash { algorithm }` — uses `serde(tag)` + `flatten` to keep the flat wire format
- Branch `compute_reveal()` by handler action: REVEAL ranges go to reveal set, HASH ranges go to commit set
- Extend `SdkProver::reveal()` to accept `Option<Commit>` and build `TranscriptCommitConfig` with per-range hash algorithms
- Update WASM `JsProver::reveal()` to accept optional commit parameter

## Design

Each `CommitRange { range, algorithm }` carries its own hash algorithm, matching tlsn-core's `TranscriptCommitConfig` which stores per-range `TranscriptCommitmentKind`. Different handlers can use different algorithms in the same proof.

`HandlerAction` is an internally-tagged enum (`#[serde(tag = "action")]`) flattened into `Handler`. This makes invalid states unrepresentable — `algorithm` only exists on the `Hash` variant and must be specified explicitly.

## Test plan

- [x] 7 new unit tests in `sdk-core/src/handler/tests.rs`
  - Mixed REVEAL/HASH handler splitting
  - Per-range algorithm propagation (SHA256, Keccak256 on different ranges)
  - HASH without algorithm fails deserialization
  - HASH action serde roundtrip with algorithm
- [x] Integration test: full prove flow with mixed reveal + commit handlers
- [ ] Plugin smoke test: ID.me plugin with `action: 'HASH'` on `full_name`

### DIFF
--- a/crates/sdk-core/src/handler/mod.rs
+++ b/crates/sdk-core/src/handler/mod.rs
@@ -18,7 +18,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     error::{Result, SdkError},
-    types::{Commit, CommitRange, Handler, HandlerType, Reveal},
+    types::{Commit, CommitRange, Handler, HandlerAction, HandlerType, Reveal},
 };
 
 /// A byte range annotated with the handler that produced it.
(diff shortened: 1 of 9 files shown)


heeckhau:
@claude review

0xtsukino (approved):

PR #1129, opened 2026-04-21
ci: add unknown target compilation check for core and attestation crates

### LINKED ISSUE #1002: Ensure `tlsn-core` and `tlsn-attestation` can compile on unknown targets
Both `tlsn-core` and `tlsn-attestation` should be able to compile without depending on syscalls (e.g. I/O, RNG).

heeckhau:
## Summary

- Add CI job `check-unknown-target-support` that verifies `tlsn-core` and `tlsn-attestation` compile with `getrandom_backend="unsupported"`, preventing regressions for unknown target support
- Add platform support documentation to both crates explaining how to compile for targets without OS-level RNG and which APIs are safe to use

Closes #1002

## Test plan

- [x] CI `check-unknown-target-support` job passes
- [x] Existing CI jobs unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


th4s (approved):
Hmm I feel like the right solution would be to have a proper `verification` crate , but given that we do not intend to continue development of `attestation` this is fine for me.

PR #1142, opened 2026-05-06
refactor: tlsn proxy API

sinui0:


crates/tlsn/src/lib.rs
```diff
@@ -99,32 +100,50 @@ pub(crate) struct TlsOutput {
...
+
+impl ProtocolConfig for ProxyTlsConfig {
+    type Commit = Proxy;
+}
+
+/// TLS commitment protocol.
+pub trait TlsCommitmentProtocol: sealed::Sealed {
+    /// Returns the kind of TLS commitment protocol.
+    fn kind(&self) -> TlsCommitKind;
+}
```
th4s: What is this for? It is unused.

th4s (approved):
lgtm, I think we do not need this extra trait?

crates/tlsn/src/lib.rs
sinui0: true, i forgot i eliminated it

PR #1146, opened 2026-05-08
Refactor `TlsTranscript`

th4s:
First draft for refactoring `TlsTranscript`.

### DIFF
--- a/crates/core/src/transcript/tls.rs
+++ b/crates/core/src/transcript/tls.rs
@@ -1,272 +1,44 @@
 //! TLS transcript.
 
-use const_oid::db::rfc5912;
-use rustls_pki_types as pki_types;
-use spki::der::{Decode, oid::ObjectIdentifier};
-
 use crate::{
-    connection::{
-        CertBinding, CertBindingV1_2, HandshakeData, KeyType, ServerEphemKey, ServerSignature,
-        SignatureAlgorithm, TlsVersion, VerifyData,
-    },
-    transcript::{Direction, Transcript},
+    connection::{CertBinding, ServerSignature, TlsVersion},
+    transcript::{Direction, Transcript, tls::builder::SfHashInput},
     webpki::CertificateDer,
 };
+
 use sha2::{Digest, Sha256};
-use tls_core::msgs::{
-    alert::AlertMessagePayload,
-    codec::{Codec, Reader},
-    enums::{
-        AlertDescription, ContentType as TlsContentType, HandshakeType, NamedGroup,
-        ProtocolVersion, SignatureScheme,
-    },
-    handshake::{HandshakeMessagePayload, HandshakePayload, KeyExchangeAlgorithm},
-    message::OpaqueMessage,
-};
 
-/// TLS record content type.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
-pub enum ContentType {
-    /// Change cipher spec protocol.
-    ChangeCipherSpec,
-    /// Alert protocol.
-    Alert,
-    /// Handshake protocol.
-    Handshake,
-    /// Application data protocol.
-    ApplicationData,
-    /// Heartbeat protocol.
-    Heartbeat,
-    /// Unknown protocol.
-    Unknown(u8),
-}
-
-impl From<ContentType> for tls_core::msgs::enums::ContentType {
-    fn from(content_type: ContentType) -> Self {
-        match content_type {
-            ContentType::ChangeCipherSpec => tls_core::msgs::enums::ContentType::ChangeCipherSpec,
-            ContentType::Alert => tls_core::msgs::enums::ContentType::Alert,
-            ContentType::Handshake => tls_core::msgs::enums::ContentType::Handshake,
-            ContentType::ApplicationData => tls_core::msgs::enums::ContentType::ApplicationData,
-            ContentType::Heartbeat => tls_core::msgs::enums::ContentType::Heartbeat,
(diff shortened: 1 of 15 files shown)


th4s:
@sinui0 This is a start, which addresses the public surface of the `TlsTranscript` type. The question is where to go from here. Is this enough, or do we want to harmonize the construction of the transcript. This would essentially mean getting rid of `TlsTranscriptBuilder::from_parts`, but entails some questions for the MPC side.

1) `MpcTlsFollower` not seeing any handshake records is ok, if we just have some optional fields on `TlsTranscript`.
2) `RecordLayer::commit` returns `Vec<Record>`. How should this be turned into raw bytes for `TlsTranscriptBuilder`. Should the record layer be able to return the raw TLS bytes and plaintext?

What' also important to note here is that the `TlsTranscript` type does not contain any correctness or authentication guarantees. Currently prover and verifier are responsible for handling it after the construction of this type and before exposing it through our api.

crates/core/src/transcript/tls/builder.rs
```diff
@@ -0,0 +1,714 @@
...
+    pub fn certificate_binding(&mut self, binding: CertBinding) -> &mut Self {
+        self.certificate_binding = Some(binding);
+        self
+    }
+
+    /// Builds a [`TlsTranscript`]. If possible will also return
+    /// [`HandshakeHashes`].
+    ///
+    /// Prefers avaialble fields, but if missing tries to parse.
+    pub fn build(self) -> Result<(TlsTranscript, Option<HandshakeHashes>), TlsTranscriptError> {
```
sinui0: Returning "handshake hashes" is a leak, unrelated to the TlsTranscript builder. `TlsTranscript` should have getters

crates/core/src/transcript/tls/builder.rs
```diff
@@ -0,0 +1,714 @@
...
+        };
+
+        Ok((transcript, handshake_hashes))
+    }
+}
+
+/// Handshake-transcript hash artifacts derived during parsing of the wire
+/// bytes.
+#[derive(Debug, Clone)]
+pub struct HandshakeHashes {
```
sinui0: This type seems unnecessary if we have getters on `TlsTranscript` which compute them.

crates/core/src/transcript/tls.rs
```diff
@@ -450,13 +121,9 @@ impl TlsTranscript {
...
-/// A TLS record.
-#[derive(Clone)]
-pub struct Record {
-    /// Sequence number.
-    pub seq: u64,
-    /// Content type.
-    pub typ: ContentType,
+/// TLS Finished Record.
+#[derive(Debug, Clone)]
+pub struct FinishedRecord {
```
sinui0: Does this type need to exist?

sinui0 (commented):
Directionally correct, I think the contract of the `TlsTranscript` type needs to be better defined.

crates/core/src/transcript/tls.rs
th4s: I first thought it was a good idea, but  now I am leaning into the other direction, especially because this creates other problems. Will probably remove it.

crates/core/src/transcript/tls.rs
```diff
@@ -314,99 +75,69 @@ impl TlsTranscript {
...
 
     /// Returns the client finished hash.
-    pub fn cf_hash(&self) -> Option<&[u8]> {
-        self.cf_hash.as_deref()
+    pub fn cf_hash(&self) -> Option<[u8; 32]> {
+        self.cf_hash.as_ref().copied()
+    }
+
+    /// Returns the session hash.
+    pub fn session_hash(&self) -> Option<[u8; 32]> {
```
sinui0, line 119: What is "session_hash"? It should be documented what this corresponds to

crates/core/src/transcript/tls.rs
```diff
@@ -314,99 +75,69 @@ impl TlsTranscript {
...
+        self.session_hash.as_ref().copied()
     }
 
-    /// Returns the server finished hash.
-    pub fn sf_hash(&self, cf_vd: &[u8; 12]) -> Option<Vec<u8>> {
+    /// Returns the server finished hash given the client finished verify
+    /// data.
+    pub fn sf_hash(&self, cf_vd: &[u8; 12]) -> Option<[u8; 32]> {
```
sinui0, line 125: Why not call `self.cf_vd()`? If it is passed here as an argument then presumably it is known and should be in `TlsTranscript` already.

crates/core/src/transcript/tls/builder.rs
```diff
@@ -0,0 +1,689 @@
...
+
+    /// Sets the certificate binding.
+    pub fn certificate_binding(&mut self, binding: CertBinding) -> &mut Self {
+        self.certificate_binding = Some(binding);
+        self
+    }
+
+    /// Builds a [`TlsTranscript`].
+    ///
+    /// Prefers avaialble fields, but if missing tries to parse.
```
sinui0:
```suggestion
    /// Prefers available fields, but if missing tries to parse.
```

sinui0 (changes requested):
Good work, definitely getting very close. The builder still needs the ability to accept partial plaintext data, e.g. the client finished verify data, being present.

crates/core/src/transcript/tls.rs
th4s: We cannot use `self.cf_vd()` because there is no plaintext available in proxy mode. The value we use as an argument is the output of the PRF.

sinui0 (approved):

PR #1147, opened 2026-05-08
refactor(tlsn): feature gate keccak and blake3

sinui0:
This PR adds feature flags to `tlsn` which allow a user to disable the keccak and blake3 hash functions. This reduces binary size.

Blake3 remains enabled by default as it is currently the default for transcript commitments and I didn't want to introduce a behavioral change.


heeckhau:
@claude review

heeckhau:
Yes for smaller binary size!
Since SHA256 is always enabled: wouldn't it make more sense to have SHA256 as default instead of BLAKE3?

th4s (approved):

sinui0:
We still want the default to be BLAKE3 as it is more efficient, that's why it is also a default enabled feature. A user who does not need it should both disable the feature and the change config.

PR #1167, opened 2026-06-11
fix: prover close connection handles on error

th4s:
### Problem
  When the MPC-TLS prover future errored, it propagated the error and dropped the   `Connected` state without closing the connection handles. Any reader parked on the `TlsConnection` — and the server socket — was never woken  and could wait indefinitely.

  ### Fix
  On `Prover::poll` returning `Err`, extract `client_io` and `server_socket` from   the `Connected` state and `close()` both before yielding the error. Closing the   prover's write half sets EOF on the peer simplex and wakes its reader, so   consumers observe closure instead of hanging. Closing is best-effort: close   errors are ignored in favor of the original error. The success path is   unchanged (handles are already closed in `Prover::poll`).

### DIFF
--- a/crates/tlsn/src/prover/future.rs
+++ b/crates/tlsn/src/prover/future.rs
@@ -1,6 +1,6 @@
 //! Future used by the [Prover].
 
-use futures::{AsyncRead, AsyncWrite, Future, FutureExt, future::FusedFuture};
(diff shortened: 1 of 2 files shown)


sinui0:
rebase pls

sinui0 (approved):

PR #1168, opened 2026-06-11
refactor: cleanup and simplify mpc-tls

th4s:
## Refactor: cleanup and simplify mpc-tls

 Follow-up to #1166 (which moved `tls-client` into `mpc-tls`): now that the TLS state
  machine and the MPC machinery live in one crate, this removes the boilerplate the old
  crate boundary forced, deletes dead fork code, and hardens a few protocol paths.

  ### Client ↔ MPC seam
  - The rustls-style incremental backend setters are replaced by a single
    `HandshakeData` handover to `prepare_encryption()` at ServerHelloDone; the
    `StartHandshake`/`SetServerRandom`/`SetServerKey` wire messages merge into one
    `ServerHello` message. No-op and write-only setters are deleted.
  - Leader methods without await points are now sync; `MpcTlsLeader` is unboxed and
    its state machine flattened (common fields grouped, slim `Ready`/`Active` phases).
  - The client-side record layer shim (seq counters, trial decryption, `DirectionState`)
    is replaced by two `encrypting`/`decrypting` flags; sequence exhaustion is impossible
    because the MPC record layer's traffic limits bound record counts far below 2^64.

  ### Dead code removal
  - Deleted unreachable rustls-fork machinery: session resumption/tickets, keylog,
    `cipher.rs`, ring-based key exchange, early-data paths, single-variant
    `Side`/`Protocol` enums, the generic `State<Data>` parameter, and ~20 unused
    pub API items. Resumption is not just dead but unsound under MPC: the client
    never learns the master secret.
  - The 6-stage typestate `ConfigBuilder` becomes `ClientConfig::new` /
    `new_with_client_auth` / `new_with_cert_resolver` (suites/groups/versions are
    fixed by what the MPC backend supports anyway).
  - `client/client/` is flattened into `client/`; `ClientConnection` and
    `ConnectionCommon` merge. TLS 1.3 message handling is retained per #1166, with
    its key-schedule stubs now local to `tls13.rs`.
  - The blanket `#[allow(dead_code, unused_imports, unreachable_pub, clippy::all)]`
    on the client module is **removed** — the whole crate is lint-clean again.
    Unused deps dropped (`sct`, `rustls-pki-types`, `aes`, `ctr`, …).

  ### Hardening
  - `opaque_into_parts` no longer panics on records shorter than nonce+tag
    (peer-triggerable DoS) — now an error, with unit tests.
  - The record `Private`/`Public` mode is derived from the content type on both
    sides instead of being trusted from the wire, removing a malicious-leader
    degree of freedom (e.g. hiding handshake plaintext from the follower).
  - `client::Error::BackendError` keeps the full `MpcTlsError` source chain.


th4s:
No regression found when benching against `main`.

sinui0 (approved):

PR #1173, opened 2026-06-30
fix(MPC deadlock): give each RCOT consumer its own instance

AdamDawidKrol:
MPC `preprocess` (`crates/tlsn/src/deps/{prover,verifier}.rs`) builds one `SharedRCOT` and hands clones to the concurrent `try_join3` branches (key-exchange / record-layer / vm). A `SharedRCOT` only performs its single batched flush once all live clones reach the flush barrier — but the branches don't all flush in the same round, so a clone whose `flush()` is parked behind a sibling branch never arrives. The barrier then waits forever (`alive=N, arrived=1`) → an intermittent deadlock during preprocessing.

**Fix:**
Give each RCOT consumer its own `SharedRCOT` instance instead of clones of a shared one, so every barrier rendezvous is independent and can't wait on a clone parked in another branch. All senders keep using the same global `delta`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1592,7 +1592,7 @@ checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
 [[package]]
 name = "clmul"
 version = "0.1.0-alpha.6"
-source = "git+https://github.com/privacy-ethereum/mpz?rev=v0.1.0-alpha.6#6ebfe619490c3155a589fc6a3be83b0976de19dc"
+source = "git+https://github.com/AdamDawidKrol/mpz?rev=c0379ef#c0379efa81cad70c21c922f65a6094da6f10c630"
 dependencies = [
  "bytemuck",
  "cfg-if",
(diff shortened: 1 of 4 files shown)


sinui0 (commented):
This introduces a subtle security flaw: the kos protocol gets instantiated with the same delta correlation and the receiver has the opportunity to re-use the same base OT setup. Our kos implementation does not do domain separation by instance, so I think this gives the receiver the opportunity to outright learn (fatal) or at least sample delta.

Two options:

1. Add instance domain separation to the kos implementation
2. Use independent instances (run base OT twice)

AdamDawidKrol:
@sinui0 Thank you for the comment! I'll answer in a few days.

AdamDawidKrol:
@sinui0 I reproduced the attack you described as a test in mpz: https://github.com/AdamDawidKrol/mpz/pull/1

But I've been trying to see how a malicious party actually accesses the data it needs, and I can't. The recovery reads the honest sender's COT keys (`keys_A ^ keys_B = delta`). In the tlsn setup the malicious party is the receiver, and its own output is the choice-independent value (delta-free) — the sender folds the receiver's choice into its keys, so flipping choices across two instances just returns the same masked value, not the complementary key. So I don't see a path where prover or notary observes the delta-bearing keys.

Am I missing something?

AdamDawidKrol:
Domain separation pushed (privacy-ethereum/mpz#446 + commit here)

themighty1:
@AdamDawidKrol , i looked at your original report. All branches at the `try_join3` juncture do flush their respective RCOT functionalities. So, i'd need some clarification in what you meant by "don't all flush in the same round".
I can also take a look at the repro branch if you have one.

Wondertan:
This also useful for us, and we would like this to be merged

PR #1176, opened 2026-08-05
feat(prover): pin TLS 1.2 for real notarization + headless notarize CLI

advatar:
TLSNotary's MPC-TLS is TLS 1.2 only; this fork's TLS 1.3 verifier path is incomplete (the notary/follower never receives the server handshake records → 'server finished was not observed'). Dual-stack servers preferred 1.3 because the prover offered it, so no real notarization completed.

Pin the prover to offer TLS 1.2 only so dual-stack servers negotiate the supported protocol. Adds a headless native prover CLI (crates/ios/examples/notarize_cli.rs) mirroring the mobile prover.

Verified: real (non-demo) notarization of live https://example.com/ through the browser-demo notary gateway over TLS 1.2 → HTTP 200 → notary-signed SignedArtifact verifies against the pinned key.

Follow-up: make the version a ProverConfig option rather than hardcoded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


advatar:
Opened accidentally against upstream by a fork-default base; withdrawing. This change targets our own fork, not tlsnotary/tlsn. Apologies for the noise.

